PR #13, opened 2020-02-18
doc: add link to latest Windows log script

PatrickLang:
The link you used is for an old script. Please use this one :)


JunSun17:
Thanks for adding the doc on windows.

PR #20, opened 2020-05-12
Fix FQDN length issues for private cluster

feiskyer:
Private cluster fqdn format is `*.UUID.privatelink.<region-name>.azmk8s.io`, use first 63 char as container name.

PR #21, opened 2020-05-19
Storage name starting and ending with "-" is invalid name.

Tatsinnit:
This fixes the trailing dash `-` in storage name issue. Associated with https://github.com/Azure/aks-periscope/pull/20 

* Its kind of undocumented, but some pseudo mention here: https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names 

cc: @feiskyer and @JunSun17


JunSun17:
LGTM.

PR #33, opened 2020-12-14
Added AAD Pod identity to connect to Storage Account

arnaud-tincelin:
- added the option to connect to the storage account using a MSI
- updated the `Export` function so that we don't recreate the same `ContainerURL` everytime we call the function
- added a `Kustomization` file to ease deployment


Tatsinnit:
Hiya @arnaud-tincelin , thank you for this. just to archive, we had chat and Arnaud will contribute to couple of things more and we can tackle this feature work which you have kindly done as a tracked item moving forward as a feature upgrade with user-scenarios etc. 

I have added `do not merge` label for now.

Once again, thank and looking forward for your contributions. 😊🙏

PR #38, opened 2021-03-30
CI and enable e-2-e basic integration testing

Tatsinnit:
This work achieves the following: (Drop one - Infra and basic test setup) 

* This work will `enable CI (continuous integration)` for this repository. 
* This work will also add the `integration test` which:
    * Test both successful and unsuccessful test by calling the `real time deploy mechanism` - 
        * This helps in testing the real thing, the deployment file has the reference of the image which gets published as the deployed daemonset periscope uses for collection of logs. 
        * **How?**
        * I have used this [Framework to create Automated Test](https://github.com/Azure-Samples/automated-test-environment-for-aks-applications) Environment using kind for testing Azure Kubernetes Service (AKS) based applications in CI (Continuous Integration) Pipeline.
            * This framework spans up a `kind` cluster and then once I am done I `stop` it. 
            * Integration test for `successful` scenario test a level deep case.
                * Check for the pods `Running status`.

**_Future Expansion work_**  (I would prefer we do this incrementally and accept current state is CI and integration test as setup)
* We can go 2-level deep to read the logs especially I think it will be more kind of replication `zipAndExport` in main file.
* Unsuccessful test future expansion: We could fail the with wrong storage and then read the pod logs and then read the output.

As future work I plan to add few unit-test for the `utils` pkg for this repo. With integration test at least this will capture any broken test and CI will help in indicating the repo health as well.

* Thanks for thoughts discussions or possibly just screenshot share: @Porges , @babbageclunk  , @matthchr, @itowlson , @davefellows , @palma21  for discussion et. al. really appreciate your time.

cc: @JunSun17 for eyes as well please. cc: @davidkydd 

Please feel free to add ideas, and things I could improve as first drop towards CI archivability. 🙏😊 

Initially it kind of started as giving it a shot but the framework and discussion really helped to put all this together.

**CI build passing badge**: 

image

**CI Build looks like this**

* [Link to last successful CI run reside here](https://github.com/Tatsinnit/aks-periscope/runs/2224131083?check_suite_focus=true) 
* [Latest build](https://github.com/Tatsinnit/aks-periscope/runs/2241471212?check_suite_focus=true)

image

image

image

image


JunSun17 (approved):
Cool, thanks a lot for working on this!

.github/workflows/ci-pipeline.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: 'Install OS Packages'
+        run: |
+          sudo apt update
+          sudo apt --fix-broken install -y
+          sudo apt install python3-pip -y
+          pip3 install setuptools
+          pip3 install pylint coverage
+        
+
+      - name: 'Install Docker'
```
Porges, line 34: I think Docker is already installed on `ubuntu-latest`. You might not need this, unless you need the latest version?

tests/aksperiscopeintegration_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	output, err := utils.RunCommandOnContainer("kubectl", "apply", "-f", "../deployment/aks-periscope.yaml", "--validate=false")
+	actual := successfulRun(t, output, err)
+
+	if actual != expectation {
+		t.Errorf("Expected successful return %v but got %v", expectation, actual)
+	}
+}
+
+func TestEndToEndIntegrationUnsuccessFulCase(t *testing.T) {
```
Porges, line 20: This seems to be identical to the success case?

tests/aksperiscopeintegration_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+func TestEndToEndIntegrationSuccessCase(t *testing.T) {
+	expectation := true
+
+	output, err := utils.RunCommandOnContainer("kubectl", "apply", "-f", "../deployment/aks-periscope.yaml", "--validate=false")
+	actual := successfulRun(t, output, err)
+
+	if actual != expectation {
+		t.Errorf("Expected successful return %v but got %v", expectation, actual)
+	}
```
babbageclunk, line 17: At this point the test knows that the deployment yaml was accepted - I guess the next steps would be to wait for the pod to run and then check for the output logs?

babbageclunk (commented):
Looks good so far! Are you adding more checks to the tests to look at the generated logs?

tests/aksperiscopeintegration_test.go
Tatsinnit: :bulb: thank you for  message, yeah, this all will change soon thank @Porges got intruding me to `omega` :)

PR #46, opened 2021-04-23
Removed vendor directory

arnaud-tincelin:
Removed the vendor directory and fixed build errors due to new package versions


JunSun17:
@arnaud-tincelin Why do you remove the vendor directory? It looks will impact dependency and also some license fles.

arnaud-tincelin:
Because I am relying on go modules. Dependencies list & versions can be found in the go.mod file.

guilhem:
Current `/vendor` is totally outdated (last commit 2y ago).
For information, current `/vendor` should look like this: https://github.com/arnaud-tincelin/aks-periscope/commit/e5b8f754e84a827ace0598a28711a3610f45ec6c (199 files change with 221,385 additions and 3,712 deletions).

It shows us that this folder wasn't used and should be a problem for anyone relying on it.

JunSun17:
@Tatsinnit This looks good to me, can you hep to verify the new vendoring works as expected? Thanks!

Tatsinnit:
Sound a good and thanks @JunSun17 , indeed, @arnaud-tincelin and I have discussed this and this will only be merged once we have tested that the local image worked well. Thanks 🙏

arnaud-tincelin:
Closed in favor of https://github.com/Azure/aks-periscope/pull/53

Tatsinnit:
Cool, @arnaud-tincelin, please let me know if you have a test image with these changes and I can also give a good test for existing changes and impact on current functionality. Thank you.

go.sum
```diff
@@ -2,88 +2,87 @@ github.com/Azure/azure-pipeline-go v0.2.1 h1:OLBdZJ3yvOn2MezlWvbrBMTEUQC72zAftRZ
 github.com/Azure/azure-pipeline-go v0.2.1/go.mod h1:UGSo8XybXnIGZ3epmeBw7Jdz+HiUVpqIlpz/HKHylF4=
```
Tatsinnit, line 2: 💡 I think there will be required changes in like: `builder/Dockerfile` : https://github.com/Azure/aks-periscope/blob/master/builder/Dockerfile 

I think either if we are keeping `mode=vendor` then we might have to add `go mod vendor` or  change line 5 : https://github.com/Azure/aks-periscope/blob/master/builder/Dockerfile#L5 to `-mod=mod`

Tatsinnit (changes requested):
💡 Added minor comment because when I was trying tp publish my local image it error'd out due to `-mod=vendor` in the line 5 for `build/dockerfile` 

* I will test it today and you can do the same so that we have made sure existing functionality all behaved the same way.

Thank you so much. 🙏

go.sum
arnaud-tincelin: I've added a new commit about the dockerfile, please let me know

Tatsinnit: All good, I will try it out as well. 

FYI: I did added PR for your fork for the changes I recommended to make it easy 🙏 https://github.com/arnaud-tincelin/aks-periscope/pull/2

README.md
```diff
@@ -121,24 +121,6 @@ After export, they will also be stored in Azure Blob Service under a container w
 
 Alternatively, AKS Periscope can be deployed directly with `kubectl`. See instructions in [Appendix].
 
-
-# Programming Guide
```
Tatsinnit, line 125: 💡 Suggestion: instead of completely removing can we please add more generic instruction for newbie users please.

* Key aspect for this article if a playing ground to get folks who like to develop on this some basic instructions to:
    * How to build locally.
    * How to test their local changes.

Might be we could simplify to much basic instructions.

go.mod
```diff
@@ -1,16 +1,11 @@
 module github.com/Azure/aks-periscope
 
+go 1.15
+
 require (
 	github.com/Azure/azure-storage-blob-go v0.7.0
-	github.com/Microsoft/go-winio v0.4.13 // indirect
```
Tatsinnit, line 5: 💡 All these were unused modules I presume?

Tatsinnit (changes requested):
💡 I would do a little ping to @JunSun17 for his final eyes, beside 2 key points rest looks good.

I will test it at my end as well. Thanks.

go.mod
arnaud-tincelin: I suppose, I let `go mod tidy` do its job 🤷‍♂️

README.md
arnaud-tincelin: In my opinion the Dockerfile (and all code in general) is the best documentation. Documentation is dangerous as it may become a pain to maintain.
I'm not anti-documentation but I believe we don't want to explain the very basic usage of `go build` and `go run` I think ^.^

To help test local changes, I agree we could do better by adding some tests :)

Tatsinnit (approved):

PR #48, opened 2021-04-28
feat: add OSM logs collector

SanyaKochhar:
This PR adds collectors for [OSM](https://github.com/openservicemesh/osm) and [SMI](https://smi-spec.io/). SMI collector can be broken out into a separate PR as well.

* Flags: in `aks-periscope.yaml`, setting flags to `OSM` turns on both osm and smi collectors. `SMI` flag turns on just smi collector
* `osm_collector` collects the following for each Open Service Mesh present on the cluster:
  * ground truth - list/configs of all resources and mutating and validating webhook configurations
  * for each osm control plan namespace - pod configs and logs of OSM controller and OSM injector pods
  * for each namespace monitored by osm: metadata, services, endpoints, service accounts, configmaps, etc.
* [moving this into a separate PR] `smi_collector` collects the traffic targets, traffic splits, httproutegroups, tcproutes and udproutes for smi policies being applied.
  * this collector can be used by any service mesh, not just osm

### DIFF
--- a/cmd/aks-periscope/aks-periscope.go
+++ b/cmd/aks-periscope/aks-periscope.go
@@ -22,7 +22,7 @@ func main() {
 		log.Fatalf("Failed to create CRD: %v", err)
 	}
 
-	clusterType := os.Getenv("CLUSTER_TYPE")
+	collectorList := strings.Fields(os.Getenv("COLLECTOR_LIST"))
 
 	// Copies self-signed cert information to container if application is running on Azure Stack Cloud.
 	// We need the cert in order to communicate with the storage account.
@@ -43,24 +43,25 @@ func main() {
 	kubeletCmdCollector := collector.NewKubeletCmdCollector(exporter)
 	systemPerfCollector := collector.NewSystemPerfCollector(exporter)
 	helmCollector := collector.NewHelmCollector(exporter)
+	osmCollector := collector.NewOsmCollector(exporter)
 
-	if strings.EqualFold(clusterType, "connectedCluster") {
-		collectors = append(collectors, containerLogsCollector)
-		collectors = append(collectors, dnsCollector)
+	collectors = append(collectors, containerLogsCollector)
+	collectors = append(collectors, dnsCollector)
+	collectors = append(collectors, kubeObjectsCollector)
+	collectors = append(collectors, networkOutboundCollector)
+
+	if contains(collectorList, "connectedCluster") {
 		collectors = append(collectors, helmCollector)
-		collectors = append(collectors, kubeObjectsCollector)
-		collectors = append(collectors, networkOutboundCollector)
-
 	} else {
-		collectors = append(collectors, containerLogsCollector)
-		collectors = append(collectors, dnsCollector)
-		collectors = append(collectors, kubeObjectsCollector)
-		collectors = append(collectors, networkOutboundCollector)
 		collectors = append(collectors, systemLogsCollector)
 		collectors = append(collectors, ipTablesCollector)
 		collectors = append(collectors, nodeLogsCollector)
 		collectors = append(collectors, kubeletCmdCollector)
 		collectors = append(collectors, systemPerfCollector)
+	}
+
+	if contains(collectorList, "OSM") {
+		collectors = append(collectors, osmCollector)
 	}
 
 	collectorGrp := new(sync.WaitGroup)
@@ -157,4 +158,13 @@ func zipAndExport(exporter interfaces.Exporter) error {
 	// TODO: remove this //nolint comment once the select{} has been removed
 	//nolint:govet
 	return nil
+}
+
+func contains(flagsList []string, flag string) bool {
+	for _, f := range flagsList {
+		if strings.EqualFold(f, flag) {
+			return true
+		}
+	}
+	return false
 }
--- a/deployment/aks-periscope.yaml
+++ b/deployment/aks-periscope.yaml
@@ -17,11 +17,17 @@ metadata:
   name: aks-periscope-role
 rules:
 - apiGroups: ["","metrics.k8s.io"]
-  resources: ["pods", "nodes", "secrets"]
-  verbs: ["get", "watch", "list"]
+  resources: ["pods", "pods/portforward", "nodes", "secrets"]
+  verbs: ["get", "watch", "list", "create"]
 - apiGroups: ["aks-periscope.azure.github.com"]
   resources: ["diagnostics"]
   verbs: ["get", "watch", "list", "create", "patch"]
+- apiGroups: ["admissionregistration.k8s.io"]
+  resources: ["mutatingwebhookconfigurations", "validatingwebhookconfigurations"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["apiextensions.k8s.io"]
+  resources: ["customresourcedefinitions"]
+  verbs: ["get", "list", "watch"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
@@ -83,7 +89,7 @@ spec:
         - configMapRef:
             name: nodelogs-config
         - configMapRef:
-            name: clustertype-config
+            name: collectors-config
         - secretRef:
             name: azureblob-secret
         volumeMounts:
@@ -136,13 +142,13 @@ metadata:
 data:
   DIAGNOSTIC_NODELOGS_LIST: /var/log/azure/cluster-provision.log /var/log/cloud-init.log
 ---
-apiVersion: v1 
-kind: ConfigMap 
+apiVersion: v1
+kind: ConfigMap
 metadata:
-  name: clustertype-config
+  name: collectors-config
   namespace: aks-periscope
 data:
-  CLUSTER_TYPE: "managedCluster" # <custom flag>
+  COLLECTOR_LIST: managedCluster # <custom flag>
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
--- a/pkg/collector/collector.go
+++ b/pkg/collector/collector.go
@@ -1,6 +1,11 @@
 package collector
 
-import "github.com/Azure/aks-periscope/pkg/interfaces"
+import (
+	"path/filepath"
+
+	"github.com/Azure/aks-periscope/pkg/interfaces"
+	"github.com/Azure/aks-periscope/pkg/utils"
+)
 
 // Type defines Collector Type
 type Type int
@@ -22,6 +27,8 @@ const (
 	NetworkOutbound
 	// NodeLogs defines NodeLogs Collector Type
 	NodeLogs
+	// Osm defines Open Service Mesh Collector Type
+	Osm
 	// SystemLogs defines SystemLogs Collector Type
 	SystemLogs
 	// SystemPerf defines SystemPerf Collector Type
@@ -30,7 +37,7 @@ const (
 
 // Name returns type name
 func (t Type) name() string {
-	return [...]string{"dns", "containerlogs", "helm", "iptables", "kubeletcmd", "kubeobjects", "networkoutbound", "nodelogs", "systemlogs", "systemperf"}[t]
+	return [...]string{"dns", "containerlogs", "helm", "iptables", "kubeletcmd", "kubeobjects", "networkoutbound", "nodelogs", "osm", "systemlogs", "systemperf"}[t]
 }
 
 // BaseCollector defines Base Collector
@@ -66,6 +73,31 @@ func (b *BaseCollector) Export() error {
 	if b.exporter != nil {
 		return b.exporter.Export(b.collectorFiles)
 	}
+
+	return nil
+}
+
+// CollectKubectlOutputToCollectorFiles collects output of a given kubectl command to a file.
+// Returns kubectl's stderr output if stdout output is empty.
+func (b *BaseCollector) CollectKubectlOutputToCollectorFiles(rootPath string, fileName string, kubeCmds []string) error {
+	outputStreams, err := utils.RunCommandOnContainerWithOutputStreams("kubectl", kubeCmds...)
+	if err != nil {
+		return err
+	}
+
+	// If kubectl stdout output is empty, i.e., there is no resource of this type within the cluster
+	// the absence of this resource is logged in the file with the relevant message from stderr (Ex: "No resource found...").
+	output := outputStreams.Stdout
+	if len(output) == 0 {
+		output = outputStreams.Stderr
+	}
+
+	resourceFile := filepath.Join(rootPath, fileName)
+	if err = utils.WriteToFile(resourceFile, output); err != nil {
+		return err
+	}
+
+	b.AddToCollectorFiles(resourceFile)
 
 	return nil
 }
--- a/pkg/collector/osm_collector.go
+++ b/pkg/collector/osm_collector.go
@@ -1,0 +1,189 @@
+package collector
+
+import (
+	"log"
+	"path/filepath"
+	"regexp"
+
+	"github.com/Azure/aks-periscope/pkg/interfaces"
+	"github.com/Azure/aks-periscope/pkg/utils"
+)
+
+// OsmCollector defines an Osm Collector struct
+type OsmCollector struct {
+	BaseCollector
+}
+
+var _ interfaces.Collector = &OsmCollector{}
+
+// NewOsmCollector is a constructor
+func NewOsmCollector(exporter interfaces.Exporter) *OsmCollector {
+	return &OsmCollector{
+		BaseCollector: BaseCollector{
+			collectorType: Osm,
+			exporter:      exporter,
+		},
+	}
+}
+
+// Collect implements the interface method
+func (collector *OsmCollector) Collect() error {
+	// Get all OSM deployments in order to collect information for various resources across all meshes in the cluster
+	meshList, err := utils.GetResourceList([]string{"get", "deployments", "--all-namespaces", "-l", "app=osm-controller", "-o", "jsonpath={..meshName}"}, " ")
+	if err != nil {
+		return err
+	}
+
+	// Directory where OSM logs will be written to
+	rootPath, err := utils.CreateCollectorDir(collector.GetName())
+	if err != nil {
+		return err
+	}
+
+	for _, meshName := range meshList {
+		meshRootPath := filepath.Join(rootPath, "mesh_"+meshName)
+
+		monitoredNamespaces, err := utils.GetResourceList([]string{"get", "namespaces", "--all-namespaces", "-l", "openservicemesh.io/monitored-by=" + meshName, "-o", "jsonpath={..name}"}, " ")
+		if err != nil {
+			log.Printf("Failed to find any namespaces monitored by OSM named '%s': %+v\n", meshName, err)
+		}
+		controllerNamespaces, err := utils.GetResourceList([]string{"get", "deployments", "--all-namespaces", "-l", "app=osm-controller,meshName=" + meshName, "-o", "jsonpath={..metadata.namespace}"}, " ")
+		if err != nil {
+			log.Printf("Failed to find controller namespace(s) for OSM named '%s': %+v\n", meshName, err)
+		}
+		callNamespaceCollectors(collector, monitoredNamespaces, controllerNamespaces, meshRootPath, meshName)
+		collectGroundTruth(collector, meshRootPath, meshName)
+	}
+	return nil
+}
+
+// callNamespaceCollectors calls functions to collect data for osm-controller namespace and namespaces monitored by a given mesh
+func callNamespaceCollectors(collector *OsmCollector, monitoredNamespaces []string, controllerNamespaces []string, rootPath string, meshName string) {
+	for _, namespace := range monitoredNamespaces {
+		namespaceRootPath := filepath.Join(rootPath, "namespace_"+namespace)
+		if err := collectDataFromEnvoys(collector, namespaceRootPath, namespace); err != nil {
+			log.Printf("Failed to collect Envoy configs in OSM monitored namespace %s: %+v", namespace, err)
+		}
+		collectNamespaceResources(collector, namespaceRootPath, namespace)
+	}
+	for _, namespace := range controllerNamespaces {
+		namespaceRootPath := filepath.Join(rootPath, "controller_namespace_"+namespace)
+		if err := collectPodLogs(collector, namespaceRootPath, namespace); err != nil {
+			log.Printf("Failed to collect pod logs for controller namespace %s: %+v", namespace, err)
+		}
+		collectNamespaceResources(collector, namespaceRootPath, namespace)
+	}
+}
+
+// collectNamespaceResources collects information about general resources in a given namespace
+func collectNamespaceResources(collector *OsmCollector, rootPath string, namespace string) {
+	if err := collectPodConfigs(collector, rootPath, namespace); err != nil {
+		log.Printf("Failed to collect pod configs for ns %s: %+v", namespace, err)
+	}
+
+	var namespaceResourcesMap = map[string][]string{
+		"metadata.json":             {"get", "namespaces", namespace, "-o", "jsonpath={..metadata}", "-o", "json"},
+		"services_list.tsv":         {"get", "services", "-n", namespace, "-o", "wide"},
+		"services.json":             {"get", "services", "-n", namespace, "-o", "json"},
+		"endpoints_list.tsv":        {"get", "endpoints", "-n", namespace, "-o", "wide"},
+		"endpoints.json":            {"get", "endpoints", "-n", namespace, "-o", "json"},
+		"configmaps_list.tsv":       {"get", "configmaps", "-n", namespace, "-o", "wide"},
+		"configmaps.json":           {"get", "configmaps", "-n", namespace, "-o", "json"},
+		"ingresses_list.tsv":        {"get", "ingresses", "-n", namespace, "-o", "wide"},
+		"ingresses.json":            {"get", "ingresses", "-n", namespace, "-o", "json"},
+		"service_accounts_list.tsv": {"get", "serviceaccounts", "-n", namespace, "-o", "wide"},
+		"service_accounts.json":     {"get", "serviceaccounts", "-n", namespace, "-o", "json"},
+		"pods_list.tsv":             {"get", "pods", "-n", namespace, "-o", "wide"},
+	}
+	for fileName, kubeCmds := range namespaceResourcesMap {
+		if err := collector.CollectKubectlOutputToCollectorFiles(rootPath, fileName, kubeCmds); err != nil {
+			log.Printf("Failed to collect %s in OSM monitored namespace %s: %+v", fileName, namespace, err)
+		}
+	}
+}
+
+// collectPodConfigs collects configs for pods in given namespace
+func collectPodConfigs(collector *OsmCollector, rootPath string, namespace string) error {
+	rootPath = filepath.Join(rootPath, "pod_configs")
+	pods, err := utils.GetResourceList([]string{"get", "pods", "-n", namespace, "-o", "jsonpath={..metadata.name}"}, " ")
+	if err != nil {
+		return err
+	}
+	for _, podName := range pods {
+		kubeCmds := []string{"get", "pods", "-n", namespace, podName, "-o", "json"}
+		if err := collector.CollectKubectlOutputToCollectorFiles(rootPath, podName+".json", kubeCmds); err != nil {
+			log.Printf("Failed to collect config for pod %s in OSM monitored namespace %s: %+v", podName, namespace, err)
+		}
+	}
+	return nil
+}
+
+// collectDataFromEnvoys collects Envoy proxy config for pods in monitored namespace: port-forward and curl config dump
+func collectDataFromEnvoys(collector *OsmCollector, rootPath string, namespace string) error {
+	pods, err := utils.GetResourceList([]string{"get", "pods", "-n", namespace, "-o", "jsonpath={..metadata.name}"}, " ")
+	if err != nil {
+		return err
+	}
+	for _, podName := range pods {
+		pid, err := utils.RunBackgroundCommand("kubectl", "port-forward", podName, "-n", namespace, "15000:15000")
+		if err != nil {
+			log.Printf("Failed to collect Envoy config for pod %s in OSM monitored namespace %s: %+v", podName, namespace, err)
+			continue
+		}
+
+		envoyQueries := [5]string{"config_dump", "clusters", "listeners", "ready", "stats"}
+		for _, query := range envoyQueries {
+			responseBody, err := utils.GetUrlWithRetries("http://localhost:15000/"+query, 5)
+			if err != nil {
+				log.Printf("Failed to collect Envoy %s for pod %s in OSM monitored namespace %s: %+v", query, podName, namespace, err)
+				continue
+			}
+			// Remove certificate secrets from Envoy config i.e., "inline_bytes" field from response
+			re := regexp.MustCompile("(?m)[\r\n]+^.*inline_bytes.*$")
+			secretRemovedResponse := re.ReplaceAllString(string(responseBody), "---redacted---")
+
+			fileName := query + "_" + podName + ".txt"
+			resourceFile := filepath.Join(rootPath, "envoy_data", fileName)
+			if err = utils.WriteToFile(resourceFile, secretRemovedResponse); err != nil {
+				log.Printf("Failed to write to file: %+v", err)
+				continue
+			}
+			collector.AddToCollectorFiles(resourceFile)
+		}
+		if err = utils.KillProcess(pid); err != nil {
+			log.Printf("Failed to kill process: %+v", err)
+			continue
+		}
+	}
+	return nil
+}
+
+// collectPodLogs collects logs of every pod in a given namespace
+func collectPodLogs(collector *OsmCollector, rootPath string, namespace string) error {
+	rootPath = filepath.Join(rootPath, "pod_logs")
+	pods, err := utils.GetResourceList([]string{"get", "pods", "-n", namespace, "-o", "jsonpath={..metadata.name}"}, " ")
+	if err != nil {
+		return err
+	}
+	for _, podName := range pods {
+		if err := collector.CollectKubectlOutputToCollectorFiles(rootPath, podName+".log", []string{"logs", "-n", namespace, podName}); err != nil {
+			log.Printf("Failed to collect logs for pod %s: %+v", podName, err)
+		}
+	}
+	return nil
+}
+
+// collectGroundTruth collects ground truth on resources in given mesh
+func collectGroundTruth(collector *OsmCollector, rootPath string, meshName string) {
+	var groundTruthMap = map[string][]string{
+		"all_resources_list.tsv":                 {"get", "all", "--all-namespaces", "-l", "app.kubernetes.io/instance=" + meshName, "-o", "wide"},
+		"all_resources_configs.json":             {"get", "all", "--all-namespaces", "-l", "app.kubernetes.io/instance=" + meshName, "-o", "json"},
+		"mutating_webhook_configurations.json":   {"get", "MutatingWebhookConfiguration", "--all-namespaces", "-l", "app.kubernetes.io/instance=" + meshName, "-o", "json"},
+		"validating_webhook_configurations.json": {"get", "ValidatingWebhookConfiguration", "--all-namespaces", "-l", "app.kubernetes.io/instance=" + meshName, "-o", "json"},
+	}
+	for fileName, kubeCmds := range groundTruthMap {
+		if err := collector.CollectKubectlOutputToCollectorFiles(rootPath, fileName, kubeCmds); err != nil {
+			log.Printf("Failed to collect %s for OSM: %+v", fileName, err)
+		}
+	}
+}
--- a/pkg/utils/helper.go
+++ b/pkg/utils/helper.go
@@ -7,10 +7,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 )
 
 const (
@@ -29,6 +31,11 @@ type Azure struct {
 // AzureStackCloud defines Azure Stack Cloud configuration
 type AzureStackCloud struct {
 	StorageEndpointSuffix string `json:"storageEndpointSuffix"`
+}
+
+type CommandOutputStreams struct {
+	Stdout string
+	Stderr string
 }
 
 // IsAzureStackCloud returns true if the application is running on Azure Stack Cloud
@@ -121,24 +128,80 @@ func RunCommandOnHost(command string, arg ...string) (string, error) {
 	return string(out), nil
 }
 
-// RunCommandOnContainer runs a command on container system
-func RunCommandOnContainer(command string, arg ...string) (string, error) {
+// RunCommandOnContainerWithOutputStreams runs a command on container system and returns both the stdout and stderr output streams
+func RunCommandOnContainerWithOutputStreams(command string, arg ...string) (CommandOutputStreams, error) {
 	cmd := exec.Command(command, arg...)
 
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	err := cmd.Run()
+	outputStreams := CommandOutputStreams{stdout.String(), stderr.String()}
+
+	if err != nil {
+		return outputStreams, fmt.Errorf("Fail to run command in container: %s", fmt.Sprint(err)+": "+stderr.String())
+	}
+
+	return outputStreams, nil
+}
+
+// RunCommandOnContainer  runs a command on container system and returns the stdout output stream
+func RunCommandOnContainer(command string, arg ...string) (string, error) {
+	outputStreams, err := RunCommandOnContainerWithOutputStreams(command, arg...)
+	return outputStreams.Stdout, err
+}
+
+// RunBackgroundCommand starts running a command on a container system in the background and returns its process ID
+func RunBackgroundCommand(command string, arg ...string) (int, error) {
+	cmd := exec.Command(command, arg...)
 	var out bytes.Buffer
 	var stderr bytes.Buffer
 	cmd.Stdout = &out
 	cmd.Stderr = &stderr
-	err := cmd.Run()
+	err := cmd.Start()
 	if err != nil {
-		return "", fmt.Errorf("Fail to run command in container: %s", fmt.Sprint(err)+": "+stderr.String())
+		return 0, fmt.Errorf("Start background command in container exited with message %s: %w", stderr.String(), err)
 	}
+	return cmd.Process.Pid, nil
+}
 
-	return out.String(), nil
+// Finds and kills a process with a given process ID
+func KillProcess(pid int) error {
+	process, err := os.FindProcess(pid)
+	if err != nil {
+		return fmt.Errorf("Find process with pid %d to kill: %w", pid, err)
+	}
+	if err := process.Kill(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Tries to issue an HTTP GET request up to maxRetries times
+func GetUrlWithRetries(url string, maxRetries int) ([]byte, error) {
+	retry := 1
+	for {
+		resp, err := http.Get(url)
+		if err != nil {
+			if retry == maxRetries {
+				return nil, fmt.Errorf("Max retries reached for request HTTP Get %s: %w", url, err)
+			}
+			retry++
+			time.Sleep(5 * time.Second)
+		} else {
+			defer resp.Body.Close()
+			return ioutil.ReadAll(resp.Body)
+		}
+	}
 }
 
 // WriteToFile writes data to a file
 func WriteToFile(fileName string, data string) error {
+	if err := os.MkdirAll(filepath.Dir(fileName), os.ModePerm); err != nil {
+		return fmt.Errorf("Fail to create path directories for file %s: %w", fileName, err)
+	}
 	f, err := os.Create(fileName)
 	if err != nil {
 		return fmt.Errorf("Fail to create file %s: %+v", fileName, err)
@@ -278,6 +341,23 @@ func CreateCRD() error {
 	}
 
 	return nil
+}
+
+// GetResourceList gets a list of all resources of given type in a specified namespace
+func GetResourceList(kubeCmds []string, separator string) ([]string, error) {
+	outputStreams, err := RunCommandOnContainerWithOutputStreams("kubectl", kubeCmds...)
+
+	if err != nil {
+		return nil, err
+	}
+
+	resourceList := outputStreams.Stdout
+	// If the resource is not found within the cluster, then log a message and do not return any resources.
+	if len(resourceList) == 0 {
+		return nil, fmt.Errorf("No '%s' resource found in the cluster for given kubectl command", kubeCmds[1])
+	}
+
+	return strings.Split(strings.Trim(resourceList, "\""), separator), nil
 }
 
 func writeDiagnosticCRD(crdName string) error {


pkg/collector/osmlogs_collector.go
```diff
@@ -0,0 +1,136 @@
...
+		"validatingWebhookConfigurations": []string{"get", "ValidatingWebhookConfiguration", "--all-namespaces", "--selector=app.kubernetes.io/name=openservicemesh.io", "-o=json"},
+	}
+
+	for fileName, kubeCmd := range groundTruthMap {
+		if err = collectResourceToFile(collector, rootPath, fileName, kubeCmd); err != nil {
+			fmt.Printf("Failed to collect %s for OSM: %+v", fileName, err)
+		}
+	}
+
+	// ** Collect information for various resources across all meshes in the cluster **
```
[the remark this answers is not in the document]
davidkydd: perhaps we introduce a new concept of a "meshCollector", or more generally something like a "chainableCollector", which is defined in the .yaml as a parent object under which the various collectors that act on the outputs of the "chainableCollector" are defined, along with whatever is needed to define the contract between them

davidkydd (commented):
Thanks @SanyaKochhar - very interesting! 

PS I think you accidentally included a built periscope binary in your commit, which you will probably want to revert (mentioning here as github doesn't support direct comments on binary files)

pkg/collector/osmlogs_collector.go
```diff
@@ -0,0 +1,136 @@
...
+		return err
+	}
+
+	collector.AddToCollectorFiles(resourceFile)
+
+	return nil
+}
+
+// ** Collect data for namespaces monitored by a given mesh **
+func collectDataFromNamespaces(collector *OsmLogsCollector, namespaces []string, rootPath, meshName string) error {
```
davidkydd, line 90: maybe implement just this as a "meshCollector"

pkg/collector/osmlogs_collector.go
```diff
@@ -0,0 +1,136 @@
...
+		if err = collectDataFromNamespaces(collector, namespacesInMesh, rootPath, meshName); err != nil {
+			fmt.Printf("Failed to collect data from OSM monitored namespaces: %+v", err)
+		}
+		if err = collectControllerLogs(collector, rootPath, meshName); err != nil {
+			fmt.Printf("Failed to collect OSM controller logs for mesh %s: %+v", meshName, err)
+		}
+	}
+
+	return nil
+}
```
davidkydd, line 70: perhaps we introduce a new concept of a "meshCollector", or more generally something like a "chainableCollector", which is defined in the .yaml as a parent object under which the various collectors that act on the outputs of the "chainableCollector" are defined, along with whatever is needed to define the parent -> child relationship (templating?)

pkg/collector/osmlogs_collector.go
```diff
@@ -0,0 +1,136 @@
...
+
+// ** Collect data for namespaces monitored by a given mesh **
+func collectDataFromNamespaces(collector *OsmLogsCollector, namespaces []string, rootPath, meshName string) error {
+	for _, namespace := range namespaces {
+		var namespaceResourcesMap = map[string][]string{
+			meshName + "_" + namespace + "_metadata":          []string{"get", "namespaces", namespace, "-o=jsonpath={..metadata}", "-o=json"},
+			meshName + "_" + namespace + "_services_table":    []string{"get", "services", "-n", namespace},
+			meshName + "_" + namespace + "_services_configs":  []string{"get", "services", "-n", namespace, "-o", "json"},
+			meshName + "_" + namespace + "_endpoints_table":   []string{"get", "endpoints", "-n", namespace},
+			meshName + "_" + namespace + "_endpoints_configs": []string{"get", "endpoints", "-n", namespace, "-o", "json"},
```
davidkydd, line 97: could implement using chained "resourceCollector" (namespacesInMesh) -> "resourceCollector" (namespace, services, endpoints)

pkg/collector/osmlogs_collector.go
```diff
@@ -0,0 +1,136 @@
...
+		if len(controllerInfoParts) > 0 {
+			podName := controllerInfoParts[0]
+			namespace := controllerInfoParts[1]
+			if err := collectResourceToFile(collector, rootPath, meshName+"_controller_logs_"+podName, []string{"logs", "-n", namespace, podName}); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
```
davidkydd, line 127: could implement using resourceCollector (line 112) -> podLogCollector (line 121)

pkg/collector/osmlogs_collector.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return err
+	}
+
+	// ** Collect ground truth on OSM resources **
+	var groundTruthMap = map[string][]string{
+		"allResourcesTable":               []string{"get", "all", "--all-namespaces", "--selector=app.kubernetes.io/name=openservicemesh.io", "-o=wide"},
+		"allResourcesConfigs":             []string{"get", "all", "--all-namespaces", "--selector=app.kubernetes.io/name=openservicemesh.io", "-o=json"},
+		"mutatingWebhookConfigurations":   []string{"get", "MutatingWebhookConfiguration", "--all-namespaces", "--selector=app.kubernetes.io/name=openservicemesh.io", "-o=json"},
+		"validatingWebhookConfigurations": []string{"get", "ValidatingWebhookConfiguration", "--all-namespaces", "--selector=app.kubernetes.io/name=openservicemesh.io", "-o=json"},
```
davidkydd, line 41: could implement using resourceCollector (see final comment @ line 130)

davidkydd (changes requested):
I think we should look to introduce a couple of new core Periscope concepts: resourceCollector, and either meshCollector (which defines the fixed hierarchical relationship represented here) or some more general purpose "chainable" or "heirarchical" functionality for defining how the output of one collector can feed into another. The specific object names and selector strings would then be defined in the yaml rather than baking OSM specifics into the code

pkg/collector/osmlogs_collector.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// Helper function to get all resoures of given type in the cluster
+func getResourceList(resource, label, outputFormat, separator string) ([]string, error) {
+	resourceList, err := utils.RunCommandOnContainer("kubectl", "get", resource, "--all-namespaces", "--selector", label, outputFormat)
+	if err != nil {
+		return nil, err
+	}
+	return strings.Split(strings.Trim(resourceList, "\""), separator), nil
+}
```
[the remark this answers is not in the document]
SanyaKochhar: I really like the idea of this chainable/hierarchical approach and +1 on defining values in the yaml... leaving specific commands in the code made us uncomfortable for sure. The reason we haven't yet been able to move these out into a yaml is that some of the values in most commands are variable (eg: they use the meshName or specific namespace). But with some more work to template resource collection I think it should be possible!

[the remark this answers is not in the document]
SanyaKochhar: Would the goal be to make this general for other collectors as well?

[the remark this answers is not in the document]
davidkydd: Yep! Your current approach makes sense, first step get down a MVP / put requirements into code then can work on making it more generalizable. Probably wouldn't try make all collectors chainable but define a new IChainableCollector interface and those collectors for which it makes sense can implement (e.g. probably not nodeLogs collector).

I'm now thinking it might make more sense if the proposed "resourceCollector" is implemented via extension of the capabilities of the current "KubeObjects" collector instead, as these are equivalent in intent.

I'm leaning towards not committing to chainable collector functionality right now as there is a lot to work out there, but I think we can quite easily refactor what you have to push the specifics out to the yaml, and add a new general ServiceMeshCollector to consume it, I had a quick go at moving the specifics out to the yaml and this is what I have (apologies for formatting, and I think its probably not valid yaml but hopefully you get the idea):

---
apiVersion: v1 
kind: ConfigMap 
metadata:
  name: kubeobjects-config
  namespace: aks-periscope
data:
  DIAGNOSTIC_KUBEOBJECTS_LIST: kube-system/pod kube-system/service kube-system/deployment #periscope default
  
  diagnostic-kubeobjects.properties |                                        # add support for consuming this kubeobject definition format in kubeobject collector
    all-resources-table.kubeobject=all
    all-resources-table.namespace=all
    all-resources-table.selector=app.kubernetes.io/name=openservicemesh.io
    all-resources-table.output=wide
    
    all-resources-configs.kubeobject=all
    all-resources-configs.namespace=all
    all-resources-configs.selector=app.kubernetes.io/name=openservicemesh.io
    all-resources-configs.output=json
    
    mutating-webhook-configurations.kubeobject=MutatingWebhookConfiguration
    mutating-webhook-configurations.namespace=all
    mutating-webhook-configurations.selector=app.kubernetes.io/name=openservicemesh.io
    mutating-webhook-configurations.output=json
    
    validating-webhook-configurations.kubeobject=all
    validating-webhook-configurations.namespace=all
    validating-webhook-configurations.selector=app.kubernetes.io/name=openservicemesh.io
    validating-webhook-configurations.output=json
---
apiVersion: v1 
kind: ConfigMap 
metadata:
  name: servicemesh-config
  namespace: aks-periscope
data:
  DIAGNOSTIC_SERVICEMESH_LIST: osm-controller                              # new collector type, collector defines logical steps and consumes configmap 
  
  osm-controller.servicemesh.properties: |                                 #prefix matches the name of one of the entries in DIAGNOSTIC_SERVICEMESH_LIST
    namespaces.selector=openservicemesh.io/monitored-by={{meshName}}
  
  osm-controller.diagnostic-kubeobjects.properties |                       #servicemesh collector runs the logic for gathering the list of meshes, and list of namespaces that each mesh is monitoring (using above servicemesh.properties.namespaces.selector) 
    metadata.kubeobject=namespaces                                         #then for each namespace it runs this collection of "kubeobjects" collectors 
    metadata.selector=jsonpath={..metadata}
    metadata.output=json
    
    services_table.kubeobject=services
    services_configs.kubeobject=services
    services_configs.output=json
    
    endpoints_table.kubeobject=endpoints
    endpoints_configs.kubeobject=endpoints
    endpoints_configs.output=json
    
    configmaps_table.kubeobject=configmaps
    configmaps_configs.kubeobject=configmaps
    configmaps_configs.output=json
    
    ingresses_table.kubeobject=ingresses
    ingresses_configs.kubeobject=ingresses
    ingresses_configs.output=json
  
    service_accounts_table.kubeobject=serviceaccounts
    service_accounts_configs.kubeobject=serviceaccounts
    service_accounts_configs.output=json
  
  osm-controller-podLogs.properties |                                                     #podLogs could be implemented as either (1) its own discrete collector or (2)baked as part of servicemeshcollector  
    pod.selector="custom-columns=NAME:{..metadata.name},NAMESPACE:{..metadata.namespace}' #perhaps this is always the same and so we can drop it from the yaml and just implement as default behaviour - only applicable if going with (2) above
---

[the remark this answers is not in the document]
davidkydd: @Tatsinnit @itowlson any thoughts on the proposal above?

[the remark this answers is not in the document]
itowlson: @davidkydd I don't really know enough to comment in this particular case, sorry.  But if you see a use for chained collectors in the near term, then it's probably worth designing for it as long as it's not onerous.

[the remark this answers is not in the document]
davidkydd: Thanks for eyeballing anyway @itowlson, gives me confidence it's at least comprehensible. There are other pending feature requests (Arc team) so maybe we see what their requirements look like before making the call. I *think* chained collectors would be a generally useful feature but I think all kinds of stupid stuff so concrete examples help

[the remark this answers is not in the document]
Tatsinnit: I will add my 2 cents as well (before discussion moves on), I can only do full review once I have time sometime tonight:

I think its great we are discussing that aspect, but I am not entirely sure how onerous it will be, in case you have some architecture in mind then please feel free to share in a draft PR for us to dig deeper and take a look, please?

Essentially if OSM and ARC work are 2 different implementation with current architecture in place, I will recommend to treat them differently, also it is important to note, that the feature request fro ARC will eventually be a PR from ARC team just like how OSM team is doing it. We will follow similar mechanism like we do in various AKS extension et. al. that internal teams owns the implementation of their mods in the tool, we will make sure it is maintainable but the will own this.

I would encourage you to share across the what chainable collector idea you have in draft and we can see how viable it will be and how different with current architecture it will be? (new can discuss that in there)

Key things to keep in mind are:

* Are we introducing something which will become hard to maintain? (aka will be great if you share your draft implementation with the idea)
* Is current design or architecture has any cons hence the suggestions is made?
* I would keep separation of concerns intact so that the new features from any team follow a specific pattern and don't break any exist functionality.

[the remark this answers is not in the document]
itowlson: @davidkydd +1, concrete examples will be _absolutely invaluable_ in generalising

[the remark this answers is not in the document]
itowlson: and +1 to all of @Tatsinnit's observations and caveats as well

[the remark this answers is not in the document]
davidkydd: @Tatsinnit sorry I don't have any draft implementations or anything yet, just an idea at very early stages. It was prompted by the recognition that what the code in this PR is fundamentally doing is:

1. running a bunch of kubeobject collectors that also include a selector and output string (osmlogs.collector.go line 44-55)
2. running a kubeobject collector for deployments with a selector and output string (osmlogs.collector.go line 32)
2.1 for each deployment.meshName returned by the kubeobject collector in 2., run another kubeobject collector to get all the namespaces that match a selector incorporating the deployment.meshName (osmlogs.collector.go line 63)
2.1.1 for each namespace returned in 2.1, unioned with the namespaces output by another kubeobject collecter (osmlogs.collector.go line 57 & 67), run a group of kubeobject collectors to get the various types of kubeobjects we are interested in from these namespaces (the collectDataFromNamespaces func).

If, rather than doing all of this using loops in the go code, we implemented a way to chain kubeobject collectors together (so that the output of the kubeobject in 2. could act as part of the input for the kubeobject in 2.1), we could define the nesting of collectors declaratively in the yaml itself, e.g. something like:

---
apiVersion: v1 
kind: ConfigMap 
metadata:
  name: kubeobject-config
  namespace: aks-periscope
data:
    first-collector.properties |                                                   #first collector is equivalent to step 2. above
            first-collector.kubeobject=deployments                                        
            first-collector.selector=app=osm-controller
            first-collector.output=jsonpath={..meshName}
     second-collector.properties |                                            #second collector is equivalent to step 2.1 above
            dependson=first-collector
            second-collector.kubeobject=namespaces                                        
            second-collector.selector=openservicemesh.io/monitored-by="+first-collector.output
            second-collector.output=jsonpath={..name}
     third-collector.properties |                                                #third collector is equivalent to step 2.1.1 above
            dependson=second-collector
            third-collector.kubeobject=services
            third-collector.namespace=second-collector.output    
            ...... etc for each resource in collectDateFromNamespaces func, all of which depend on second-collector
            
The intended benefit is that more complicated use cases of Periscope (such as what is contained in this PR) could be supported by the building blocks of Periscope itself; and rather than requiring code changes and a whole new build and container image to be pushed and referenced, the same thing can be accomplished with just changes to the deployment yaml.

Note I am also _not_ suggesting we implement chainable collectors right now, its just an idea that seems to be a good fit for the current scenario, as well as a generally useful capability going forward (premature generalization is a thing!).

What I _am_ suggesting we do is extend the kubeobject collector just a little bit, to support selector and output strings, and to be defined as a configmap with file-like keys. Once this is in place it should be straightforward to refactor the code in this PR to implement a more generalized ServiceMeshCollector, and the OSM specific parts can be moved out to the yaml where they belong.
    
It would also be worth taking a look at a default istio example to see if this has the same general shape as OSM (in terms of the k8s object relationships that would be baked into the collector). Otherwise our ServiceMeshCollector might not be very generalizable after all! :)

Hope that make sense? Hit me up on teams or something if you prefer once you've had time to look and we can chat more.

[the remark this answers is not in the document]
Tatsinnit: @davidkydd , Take your time please and  no-rush for your Draft PR, like mentioned as this could be a great idea and we could explore this with concrete example as mentioned in comments above, that way we can explore all the key mention in here  https://github.com/Azure/aks-periscope/pull/48#discussion_r623554119  **and**  https://github.com/Azure/aks-periscope/pull/48#discussion_r623550397 

Doing your suggestions will help us to explore the idea and design impact from what current design is and it as separate draft PR help us to share thoughts and concrete generalisation example will help us to better understand. Thank you. 🙏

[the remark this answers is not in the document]
davidkydd: Sweet as, I'll have a go at the extensions to the kubeobject collector first then. @SanyaKochhar I assume there isn't a pressing need to have your current PR merged (as you can deploy a periscope build out of a private ACR)?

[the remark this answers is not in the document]
draychev: Hi @davidkydd! It would be great to augment Periscope with these changes sooner than later - Open Service Mesh is already in public preview as on AKS addon.

Once we have the prototype in this PR working, it should be straight forward to refactor to DRY it up, and later refactor again to fit the new design you are working on.

We could also wrap these changes with a feature flag to be able to continue to iterate with smaller PRs.

I propose we
1. address as many of the immediate comments to this PR as possible to make it fit the current design
2. merge the PR
3. adapt OSM + SMI code to the new model / design you are working on

Do these sound like reasonable next steps?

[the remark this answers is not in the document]
davidkydd: Hi @draychev :) sure, in that case refactoring later should be fine. 

Feature flag at a minimum would be great as test coverage is currently very limited so regressions are significant concern. Should be able to implement one by adding a new envFrom configmap defining which collectors are enabled (simple list of strings based on type names in collector.go should suffice), and then reading the config and conditionally adding only those collectors that are enabled in aks-periscope.go. Open to better ideas though :)

Other than that I believe the only blocker to merging is the removal of the aks-periscope binary at the project root. We then manually build and push a new containerImage to aksrepos.azurecr before updating the image tag ourselves in a separate PR - is that the right process @Tatsinnit?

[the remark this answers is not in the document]
Tatsinnit: Agree:

* This is the best step moving forward, and like the idea of treating any design refactor separately so that we can actually explore what the idea is in more clarity.

* I will add bunch of checklist below as separate comment and yes, at least basic checks and due-diligence to check that these changes work with OSM\SMI feature flagged and don't interfere with normal workflow.
     * Key thing is this PR **Can-Only** be merged once all those checks are satisfied. So, please about eager merges.
     * At Dev-level falls will help this code path to only execute when the conditions are met. 

* Cool, @davidkydd  As mentioned, I would recommend you can open a workitem to track this specific re-design part and self assign it to yourself please, for the thoughts and ideas you've shared for this design refactor separately, and I believe this no longer is part of this drop, so this is un-blocked now, I think if we all agree then we could mark this resolved once we have that work item open or I could do it latter and assign it to you. 
    * When re-design \ refactoring the idea: please do take care ofpoint mentioned here  https://github.com/Azure/aks-periscope/pull/48#discussion_r623554119  **and**  https://github.com/Azure/aks-periscope/pull/48#discussion_r623550397 

Please feel free to add anything else need discussion otherwise this kind of sums it all.

Thank you guys 🙏

pkg/collector/collector.go
```diff
@@ -67,3 +76,26 @@ func (b *BaseCollector) Export() error {
...
+		return err
+	}
+
+	output := outputStreams.Stdout
+	// If kubectl stdout output is empty, i.e., there is no resource of this type within the cluster, the absence of this resource is logged in the file with the relevant message from stderr (Ex: "No resource found...").
+	if len(output) == 0 {
+		output = outputStreams.Stderr
+	}
+
+	if err = utils.WriteToFile(resourceFile, output); err != nil {
```
Tatsinnit, line 94: 💡Code  thought: I really like the idea of on-failure return early.  That way the `resourceFile := filepath.Join(rootPath, fileName)` don't need to be used until the very end.

pkg/utils/helper.go
```diff
@@ -59,20 +64,29 @@ func RunCommandOnHost(command string, arg ...string) (string, error) {
 	return string(out), nil
 }
 
-// RunCommandOnContainer runs a command on container system
-func RunCommandOnContainer(command string, arg ...string) (string, error) {
+// RunCommandOnContainerWithOutputStreams runs a command on container system and returns both the stdout and stderr output streams
+func RunCommandOnContainerWithOutputStreams(command string, arg ...string) (CommandOutputStreams, error) {
```
Tatsinnit, line 68: 💡I like what you have done, thankyou. To more fully share some thoughts, this way we need to make sure all exisiting functionality are working fine, otherwise, if you find any issue latter here is my 2 cents of idea to un-block you in those scenarios: Why not create separate helper function for the specific OSM needs? Will be a good use of extending Function for the generic function of OSM/SMi specific need?

We could do this as more systematically i.e..:

* Create new helper function with this name and keep old one as it is so that old code consumes whatever it is soncusming right now.
* Once OSM\SMI code is consuming the new code then transition old soncumers into new generic function?

Feel free to add thoughts once you have tested this,

Tatsinnit (changes requested):
💡 Thank you so much for this, I have added initial  thoughts for the key things like flag passing, and what all things you could explore:

few things as a checklist to make sure you have covered existing behaviours are:

:white_check_mark: Make sure you locally run CI. (Adding more clarity: locally atleast initial infra e2-e test are passing and your forked branch ci-workflow is passing) 
:white_check_mark: Tip to locally test changes via image reside here: https://github.com/Azure/aks-periscope#programming-guide 

:white_check_mark: If you find any new functionality could be easily extracted to UT (unit test) then feel enabled to do so.
:white_check_mark: Once we have done all the above, we should have enough confidence to open PR for review. 

(At least these are my initial thoughts) I will add more in case I forget anything.

Really appreciate it. @arnaud-tincelin and @JunSun17 Feel free to correct me or add anything else.

pkg/utils/helper.go
davidkydd: _Create new helper function with this name and keep old one as it is so that old code consumes whatever it is soncusming right now._ isn't this exactly what has been done? (see line 87 of helper.go). Unit test would be nice but this change seems straightforward enough that it doesn't require it

pkg/collector/collector.go
```diff
@@ -67,3 +76,26 @@ func (b *BaseCollector) Export() error {
 
 	return nil
 }
+
+// Helper function to collect output of a given kubectl command to a file. Returns kubectl's stderr output if stdout output is empty.
+func (b *BaseCollector) collectKubeResourceToFile(rootPath, fileName string, kubeCmds []string) error {
```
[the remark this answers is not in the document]
davidkydd: I feel we can accept this as is for now and address as part of later refactor (this func is partially duplicated elsewhere anyway)

pkg/utils/helper.go
Tatsinnit: 💡  key point I was trying to bring across as idea share is to be watchful of the the old function - now consumes new function with few cabbages, hence as part of this change the it could be litte onerous to check all impact. 

Hence, if that turns out to be hard to test in any way possible feel free to add this only for use of new functionality and carefully port old functionality to use new function. Thanks.

pkg/collector/collector.go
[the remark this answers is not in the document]
Tatsinnit: 💡 please note and remember: these points are only for author to keep in mind, in case it’s super easy or it helped them in their design to be cleaner they are welcome to take onboard but beyond.

mas long as we have captured those thinking and thoughts I am happy for any moving forward decisions, changing design is much bigger refactor then moving functions between files if the underlying workflow needs it. 

Again, thought share for Original authors to way in and decide.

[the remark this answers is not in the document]
davidkydd: sure - just sharing my opinion too :)

pkg/collector/collector.go
shalier: @Tatsinnit could you expand on what you mean by that?

Tatsinnit: 💡 Code comment was to do with 2 key thing first is the readability, flow of code and second error handling. Also, I was thinking that for consistency why not like this:

* Since `resourceFile` is not used at all till it reaches this code? Also t
    * Also could entirely move to separate func with better name which will identify the exact purpose of this check and add it to the flow?

``` 
resourceFile := filepath.Join(rootPath, fileName)
err = utils.WriteToFile(resourceFile, output);
if err != nil {
      // do something
}
```

further: 

* In case you want to have bunch of error handling `err != nil` scenario you could explore common method for this particular work.

pkg/utils/helper.go
Tatsinnit: 💡 @shailer, this comment was to be careful to make sure all current functionality will get affected in what extent? (So that this could be documented) because the new way the old method signature is now: (Also, I testing will be key for this as well) 

* `RunCommandOnContainer` now consumes `RunCommandOnContainerWithOutputStreams` and new method introduced now use `commandStream`? 

If we have added\changed\introduced new log behaviours it will be key to capture those for documentation purposes please.

``` 
// RunCommandOnContainer  runs a command on container system and returns the stdout output stream
 func RunCommandOnContainer(command string, arg ...string) (string, error) {
 	outputStreams, err := RunCommandOnContainerWithOutputStreams(command, arg...)
 	return outputStreams.Stdout, err
 }
```

davidkydd: @Tatsinnit this particular change is really quite trivial and I don't think it needs any new documentation or further work in this PR...

CommandOutputStreams (which I think is what you mean by commandStream), is simply a new struct wrapping both stdOut and stdErr, defined a couple lines up in this file (helpers.go lines 14-18).

The current functionality is maintained _exactly_ by having RunCommandOnContainer only return the stdOut field of this struct.

Tatsinnit: @davidkydd, yes, I am aware. Like I have mentioned few times to you, my comments when it comes down to exploring existing functionality are purely careful and defensive, it is upto author to take the thoughts and decide the design flow, the ideas I have given it purely be cautious and aware in case anything new functionality will start appearing for existing feature.

* Basically, keep in mind and if we are adding any new thing to an existing functionality I want it to be added as simple as one liner, either as comment, for anyone else to be aware and it will help other readers.

davidkydd: I agree that they are fair comments, and show admirable intent!

Are you satisfied with @shalier's explanation of the local testing that has been performed? In which case we can look to resolve this particular thread?

pkg/collector/osmlogs_collector.go
```diff
@@ -0,0 +1,161 @@
...
+			namespace := controllerInfoParts[1]
+			if err := collector.collectKubeResourceToFile(rootPath, meshName+"_controller_logs_"+podName, []string{"logs", "-n", namespace, podName}); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// Helper function to get a list of all resoures of given type in a specified namespace
```
davidkydd, line 145: typo

cmd/aks-periscope/aks-periscope.go
```diff
@@ -41,6 +41,10 @@ func main() {
 	collectors = append(collectors, kubeletCmdCollector)
 	systemPerfCollector := collector.NewSystemPerfCollector(exporter)
 	collectors = append(collectors, systemPerfCollector)
+	osmLogsCollector := collector.NewOsmLogsCollector(exporter)
```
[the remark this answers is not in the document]
davidkydd: FYI PR out for a simple collector / diagnoser feature flag here: https://github.com/Azure/aks-periscope/pull/52

pkg/utils/helper.go
shalier: @davidkydd @Tatsinnit  SGTM to resolve this thread, thanks for the explanation!

pkg/collector/collector.go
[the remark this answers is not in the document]
shalier: @Tatsinnit @davidkydd 
We thought it would help others if we kept the function in `pkg/collector/collector.go`  since it seemed that the rest of the repo mostly does the same thing
-- collect kubectl resource and then write to file

SanyaKochhar (commented):
Thanks folks for your thorough reviews and comments. Based on this, on our end we have been tracking and working on the following tasks, do let me know if anything is missing.
* Toggle osm collector with a feature flag (to come after @davidkydd 's PR #52  is merged)
* Write unit tests if possible
* Extract methods into `helpers.go` as possible
* Implement on-failure return early

cc: @davidkydd  @Tatsinnit

pkg/collector/collector.go
shalier: @Tatsinnit  You're right `resourceFile` should be moved further down the code. I don't think we should separate the function because `utils.WriteToFile()` requires `output` as a parameter

Tatsinnit: 💡 Sure, yeah key idea is code cleanliness and readability. feel free to scope it otherwise we can do it latter, key thing is that dividing `functs` to smaller unit usually will help to see, it might helps in UTs (latter as well).

cmd/aks-periscope/aks-periscope.go
```diff
@@ -42,6 +43,21 @@ func main() {
 	systemPerfCollector := collector.NewSystemPerfCollector(exporter)
 	collectors = append(collectors, systemPerfCollector)
 
+	// Checks flags and adds specified collector if matched
+	flags := strings.Fields(os.Getenv("FLAGS_LIST"))
+	isOsmEnabled := false
```
Tatsinnit, line 48: 💡 🐛 This will introduce a bug in your implementation logic here: What I read from below code is `OSM` and `SMI` are mutually exclusive flags? (because if `isOsmEnabled` is **true**, the subsequent flag for `SMI` will not reach because `&& !isOsmEnabled`

Further: 

* `isOsmEnabled := false` and then used for `if` conditions (I think these are the only places where this car is used)

* What is the use case of list getting passed with `OSM and SMI` together? In this case your `isOsmEnabled` is set to `true` already?

deployment/aks-periscope.yaml
```diff
@@ -108,30 +119,38 @@ data:
   AZURE_BLOB_ACCOUNT_NAME: # <accountName, base64 encoded>
   AZURE_BLOB_SAS_KEY: # <saskey, base64 encoded>
 ---
-apiVersion: v1 
-kind: ConfigMap 
+apiVersion: v1
```
Tatsinnit, line 122: 💡What is the change here?

deployment/aks-periscope.yaml
```diff
@@ -108,30 +119,38 @@ data:
...
 data:
   DIAGNOSTIC_NODELOGS_LIST: /var/log/azure/cluster-provision.log /var/log/cloud-init.log
 ---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: collectorflags-config
+  namespace: aks-periscope
+data:
+  FLAGS_LIST: # enable collector: OSM, SMI
```
Tatsinnit, line 152: 💡 `ARC` PR already introduces this flag: https://github.com/Azure/aks-periscope/pull/54/files#diff-cabcaac1a231e2d530ba374ca786e837a4c7f802f98c86299b93874aaf2e78eaR138

pkg/collector/collector.go
```diff
@@ -20,6 +25,10 @@ const (
 	NetworkOutbound
 	// NodeLogs defines NodeLogs Collector Type
 	NodeLogs
+	// Osm defines Osm Collector Type
```
Tatsinnit, line 28: 💡 Will recommend to add what OSM and SMI stand for in comments will be a good place.

pkg/collector/smi_collector.go
```diff
@@ -0,0 +1,79 @@
...
+	crdList := strings.Split(strings.Trim(outputStreams.Stdout, "\""), " ")
+	// If the CRD is not found within the cluster, then log a message and do not return any resources.
+	if len(crdList) == 0 {
+		err := fmt.Errorf("No CustomResourceDefinitions resource found in the cluster.")
+		return nil, err
+	}
+
+	var smiCrdList []string
+	for _, crd := range crdList {
+		if strings.Contains(crd, "smi-spec.io") {
```
Tatsinnit, line 73: 💡 this checks the output contains `smi-spec.io` but then if it does already, why create a new list and add it to new list and return?

* I guess, I am trying to understand what is the intention of 2 list here?

pkg/collector/osm_collector.go
```diff
@@ -0,0 +1,189 @@
...
+		return err
+	}
+
+	// Directory where OSM logs will be written to
+	rootPath, err := utils.CreateCollectorDir(collector.GetName())
+	if err != nil {
+		return err
+	}
+
+	for _, meshName := range meshList {
```
Tatsinnit, line 43: 💡 What is this Range trying to  achieve in the OSM collector context here? (Reason I ask is because if I correctly understand the needs, we could simplify this)

Tatsinnit (changes requested):
💡 Thank you for this. I have not done a detailed review because there are some higher level questions this PR needs to cater.

* I think you guys work with ARC team? if yes please liason with them, as their PR will be merged soon, ARC - PR already adds a feature flag which you could use. Some of my comments are specific to how the logic is constructed around the `OSM` and `SMI` flags and the built up.
* Please remove the `aks-periscope` [Binary which is checked in with this PR](https://github.com/Azure/aks-periscope/blob/af07833f8aac5e69ce9d6197045cd3207b66386a/aks-periscope).

* Side note: Is there any internal design doc you guys have for this work? (Something as basic of what OSM collector and SMI collector aim to achieve with what commands which are run) 
    * it would help in understanding the nested for loops with in collectors as well etc.

Thank you. 🙏

pkg/collector/osm_collector.go
SanyaKochhar: This is needed in case there are multiple instances of OSM installed on the cluster. In that case we would want to collect everything for each of the OSMs

pkg/collector/smi_collector.go
SanyaKochhar: @johnsonshi

cmd/aks-periscope/aks-periscope.go
SanyaKochhar: @shalier

deployment/aks-periscope.yaml
SanyaKochhar: My guess is one of our IDEs probably was set to automatically trim whitespaces out of yamls. We can revert this change if you'd like

deployment/aks-periscope.yaml
SanyaKochhar: This is good. Left a comment on PR #54 about making the flag name generic so that we can reuse it

SanyaKochhar:
@Tatsinnit 
Here is a short design doc to explain this work. Hopefully it explains all the loops and helps with the review process. Lmk if you have any questions.
# Design

OSM work on [AKS Periscope](https://github.com/Azure/aks-periscope) aims to add two new collectors to AKS Periscope that can be optionally enabled with flags when users wish to collect OSM and SMI data. These collectors will allow users to create a snapshot of their OSM instance(s) and SMI policies in their cluster to provide to the OSM team for troubleshooting. This is especially useful so that users do not need to provide access to their cluster for troubleshooting.

## OSM  Collector
The OSM collector can be turned on using the `OSM` flag. This will automatically also turn on the [SMI collector](#smi-collector) to collect information about SMI policies being used in the cluster.

The OSM collector will look for all the OSM instances present in the cluster. For each mesh, it will collect the following information:
* Basic ground truth about the OSM instance
  - Mutating webhook configurations
  - Validating webhook configurations
  - List of all resources present with the label `app.kubernetes.io/instance=<mesh name>`
  - JSON/configs of all resources present with the label `app.kubernetes.io/instance=<mesh name>`
* Relevant data for all namespaces monitored by OSM:
  - For each pod's Envoy sidecar: config dump, clusters, listeners, ready and stats
  - Metadata, services, endpoints, ConfigMaps, ingresses and service accounts
  - List of all pods present and config of each pod
* Relevant data for all OSM controller namespaces:
  - Logs of each pod in the controller namespace (i.e., logs for `osm-controller` and `osm-injector`)
  - Metadata, services, endpoints, ConfigMaps, ingresses and service accounts
  - List of all pods present and config of each pod

## SMI Collector
The SMI collector can be enabled using the `SMI` flag to collect information about [Service Mesh Interface](https://github.com/servicemeshinterface/smi-spec) policies being used in the cluster for any service mesh, not just OSM. It is, however, automatically turned on if the OSM collector is enabled. The SMI collector collects lists and configs of the following resources:

* HTTP route groups
* TCP routes
* Traffic splits
* Traffic targets
* Traffic access
* UDP routes

## Helpers
The following helpers are needed to achieve the above work:
* `utils.GetResourceList()` expects a `kubectl` command including a resource type and namespace. Returns a list of all resources of the given type in the specified namespace
* `utils.RunCommandOnContainerWithOutputStreams()` a modified version of `utils.RunCommandOnContainer` that returns both the stdout and stderr output streams from the given command. This is used by `GetResourceList` to make sure an error is documented if a resource is not found and thus empty collector files are not erroneously created for resources that do not exist
* `utils.RunBackgroundCommand()` starts running a command on a container and returns its process ID. It does not wait for the command to return. This is used to run `kubectl port-forward` to collect Envoy information
* `utils.KillProcess()` is used to kill a process by PID (used in conjunction with background commands)
* `collector.CollectKubectlOutputToCollectorFiles()` - expects a `kubectl` command, rootpath and filename. It runs the command and saves the output to a file in that collector's file directory

cmd/aks-periscope/aks-periscope.go
shalier: >What is the use case of list getting passed with OSM and SMI together? In this case your isOsmEnabled is set to true already?

If the user accidentally sets both `OSM` and `SMI` flag, then we wanted to avoid appending the SMI collector twice to `collectors`. To prevent this I had the `isOsmEnabled` boolean, but I see there's an issue where it wouldn't catch double-appending if the user enables the flags in a different order `SMI OSM` 

If running the SMI collector twice isn't a concern we can remove it or I can work on it some more to prevent the SMI collector from running twice, lmk what you think @Tatsinnit?

SanyaKochhar: (some context to how these two flags interact) OSM applies SMI policies, so when the `OSM` collector is in use, we also want information about SMI policies being applied. However, SMI policies can be applied with other service meshes as well so the `SMI` flag ensures that users who want to collect just SMI information are able to do that regardless of which service mesh solution they are using.
> * What is the use case of list getting passed with OSM and SMI together?

Ideally this should never happen but could happen by mistake, hence the logic Shalier explained above

SanyaKochhar:
> * I think you guys work with ARC team? if yes please liason with them, as their PR will be merged soon, ARC - PR already adds a feature flag which you could use. Some of my comments are specific to how the logic is constructed around the OSM and SMI flags and the built up.

Makes sense to reuse their flag. I have left a [comment](https://github.com/Azure/aks-periscope/pull/54/files#r638060089) on their PR about using a generic flag name that would make sense for us to reuse. Of course we can also change it after their PR is merged. The comments above and design doc hopefully explain the logic/interaction behind the two flags

> * Please remove the aks-periscope Binary which is checked in with this PR.

Accidental addition, removed now!

deployment/aks-periscope.yaml
SanyaKochhar: @shalier will update this flag and logic once PR 54 is merged

Tatsinnit (changes requested):

cmd/aks-periscope/aks-periscope.go
shalier: @Tatsinnit Good point, we'll implement this way in the SMI pr

pkg/collector/smi_collector.go
SanyaKochhar: SMI stuff is being moved into a follow-up PR

cmd/aks-periscope/aks-periscope.go
```diff
@@ -35,19 +35,24 @@ func main() {
...
 	helmCollector := collector.NewHelmCollector(exporter)
+	osmCollector := collector.NewOsmCollector(exporter)
+
+	collectors = append(collectors, containerLogsCollector)
+	collectors = append(collectors, dnsCollector)
+	collectors = append(collectors, kubeObjectsCollector)
+	collectors = append(collectors, networkOutboundCollector)
+
+	connectedClusterEnabled := false
+	for _, flag := range collectorFlags {
```
arnaud-tincelin, line 46: Should it rather be a switch case instead of a for?

deployment/aks-periscope.yaml
```diff
@@ -134,13 +140,13 @@ metadata:
...
-kind: ConfigMap 
+apiVersion: v1
+kind: ConfigMap
 metadata:
-  name: clustertype-config
+  name: collectorflags-config
   namespace: aks-periscope
 data:
-  CLUSTER_TYPE: "managedCluster" # <custom flag>
+  COLLECTOR_FLAGS_LIST: # <custom flag> connectedCluster OSM
```
arnaud-tincelin, line 149: Could COLLECTOR_LIST be a better name?

Tatsinnit (changes requested):
💡 Thank you and looks much better, and great to see how everything has come together, few things I have mentioned in my review.

As original logic with regards to the nitty-gritty for the specific commands for `ism` collector I believe that you should make sure that you guys are satisfied.

- [ ] Please do test your changes with local image and I will try as well.

Thank you 🙏

pkg/utils/helper.go
```diff
@@ -59,24 +67,82 @@ func RunCommandOnHost(command string, arg ...string) (string, error) {
...
+		return fmt.Errorf("Failed to find process with pid %d to kill: ", pid, fmt.Sprint(err))
+	}
+	if err := process.Kill(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Tries to issue an HTTP GET request up to maxRetries times
+func GetUrlWithRetries(url string, maxRetries int) ([]byte, error) {
```
arnaud-tincelin, line 122: Same stuff, less code :) 
```go
func GetUrlWithRetries(url string, maxRetries int) ([]byte, error) {
	retry := 1
	for {
		resp, err := http.Get(url)
		if err != nil {
			if retry == maxRetries {
				return nil, fmt.Errorf("max retries reached for request HTTP Get %s: %w", url, err)
			}
			time.Sleep(5 * time.Second)
		} else {
			defer resp.Body.Close()
			log.Printf("HTTP GET request succeeded for url %s", url)
			return ioutil.ReadAll(resp.Body)
		}
	}
}
```

pkg/utils/helper.go
```diff
@@ -59,24 +67,82 @@ func RunCommandOnHost(command string, arg ...string) (string, error) {
...
+	}
+	log.Printf("HTTP GET request succeeded for url %s", url)
+	defer resp.Body.Close()
+	return ioutil.ReadAll(resp.Body)
 }
 
 // WriteToFile writes data to a file
 func WriteToFile(fileName string, data string) error {
+	if err := os.MkdirAll(filepath.Dir(fileName), os.ModePerm); err != nil {
+		return fmt.Errorf("Fail to create path directories for file %s: %+v", fileName, err)
```
arnaud-tincelin, line 144: use `%w` instead of `%v` in fmt.Errorf to wrap errors

pkg/utils/helper.go
```diff
@@ -218,6 +284,24 @@ func CreateCRD() error {
...
+	outputStreams, err := RunCommandOnContainerWithOutputStreams("kubectl", kubeCmds...)
+
+	if err != nil {
+		return nil, err
+	}
+
+	resourceList := outputStreams.Stdout
+	// If the resource is not found within the cluster, then log a message and do not return any resources.
+	if len(resourceList) == 0 {
+		err := fmt.Errorf("No '%s' resource found in the cluster for given kubectl command", kubeCmds[1])
```
arnaud-tincelin, line 298: `return nil, fmt.Errorf("No '%s' resource found in the cluster for given kubectl command", kubeCmds[1])`

arnaud-tincelin (changes requested):

deployment/aks-periscope.yaml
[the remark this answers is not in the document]
shalier: @Tatsinnit just double checking are you suggesting something like this 
`COLLECTOR_FLAGS_LIST: managedCluster # <custom flag> connectedCluster OSM` or 
`COLLECTOR_FLAGS_LIST: managedCluster # <custom flag>` ?

cmd/aks-periscope/aks-periscope.go
```diff
@@ -35,19 +35,24 @@ func main() {
...
+		if strings.EqualFold(flag, "connectedCluster") {
```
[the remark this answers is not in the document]
shalier: @Tatsinnit that makes sense, I'll go ahead and implement it similar to your suggestions

pkg/collector/osm_collector.go
```diff
@@ -0,0 +1,189 @@
...
+
+		envoyQueries := [5]string{"config_dump", "clusters", "listeners", "ready", "stats"}
+		for _, query := range envoyQueries {
+			responseBody, err := utils.GetUrlWithRetries("http://localhost:15000/"+query, 5)
+			if err != nil {
+				log.Printf("Failed to collect Envoy %s for pod %s in OSM monitored namespace %s: %+v", query, podName, namespace, err)
+				continue
+			}
+			// Remove secrets from response
+			re := regexp.MustCompile("(?m)[\r\n]+^.*inline_bytes.*$")
```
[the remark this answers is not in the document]
SanyaKochhar: The secrets are the certificates being added in the Envoy config as the `inline_bytes` field. Adding documentation for this as well.

cmd/aks-periscope/aks-periscope.go
shalier: Changing the implementation similar to Tatsat's suggestion, so I don't think a switch case will be needed

pkg/utils/helper.go
SanyaKochhar: @arnaud-tincelin this looks good, just one quick question. Does `retry` increment on its own or should I add a `retry++` before `time.Sleep(5 * time.Second)`

arnaud-tincelin: @SanyaKochhar you are right, I forgot a retry++. I admit I didn't test this code :)
Additionally, I don't know if the `log.Printf("HTTP GET request succeeded for url %s", url)` line if very useful, it's up to you to keep it or not :)

Why is this retry mechanism needed? Is it normal the HTTP Get fails from time to time?

deployment/aks-periscope.yaml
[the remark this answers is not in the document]
Tatsinnit: This sounds good: `COLLECTOR_FLAGS_LIST: managedCluster # <custom flag>` that way consuming tool can do a replace for ` # <custom flag>` to whatever replacement string is.

pkg/utils/helper.go
SanyaKochhar: You're right, the extra log line made more sense with the old loop's error messaging but is redundant now. 

The retry mechanism is needed because it seems like there is sometimes a lag between when the port forwarding is started and the get request goes through, so each time the first request was failing

pkg/collector/osm_collector.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+}
+
+// ** Collects information about general resources in a given namespace **
+func collectNamespaceResources(collector *OsmCollector, rootPath, namespace string) {
+	if err := collectPodConfigs(collector, rootPath, namespace); err != nil {
+		log.Printf("Failed to collect pod configs for ns %s: %+v", namespace, err)
+	}
+
+	var namespaceResourcesMap = map[string][]string{
```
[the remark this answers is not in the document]
SanyaKochhar: Yep we have tested all of this logic with local images

Tatsinnit (approved):

Tatsinnit (approved):
👍 Thank you for this, really appreciate it. Approved it and will await till evening for few more eyes. Thank you @SanyaKochhar for testing them locally as well.

pkg/utils/helper.go
```diff
@@ -121,24 +128,80 @@ func RunCommandOnHost(command string, arg ...string) (string, error) {
...
+func RunBackgroundCommand(command string, arg ...string) (int, error) {
+	cmd := exec.Command(command, arg...)
 	var out bytes.Buffer
 	var stderr bytes.Buffer
 	cmd.Stdout = &out
 	cmd.Stderr = &stderr
-	err := cmd.Run()
+	err := cmd.Start()
+	if err != nil {
+		return 0, fmt.Errorf("Fail to start background command in container: %s", fmt.Sprint(err)+": "+stderr.String())
```
arnaud-tincelin, line 165: I think doing 2 formats and a string concatenation can be simplified :)
`fmt.Errorf("start background command in container exited with message %s: %w", stderr.String(), err))`

pkg/utils/helper.go
```diff
@@ -121,24 +128,80 @@ func RunCommandOnHost(command string, arg ...string) (string, error) {
...
+	}
+	return cmd.Process.Pid, nil
+}
+
+// Finds and kills a process with a given process ID
+func KillProcess(pid int) error {
+	process, err := os.FindProcess(pid)
 	if err != nil {
-		return "", fmt.Errorf("Fail to run command in container: %s", fmt.Sprint(err)+": "+stderr.String())
+		return fmt.Errorf("Failed to find process with pid %d to kill: %+v", pid, fmt.Sprint(err))
```
arnaud-tincelin, line 174: use %w to wrap error:
`fmt.Errorf("find process with pid %d to kill: %w", pid, err)`

SanyaKochhar:
@arnaud-tincelin thanks for your comments, I've just resolved them. Would appreciate if you could re-review when you get a chance

PR #51, opened 2021-05-05
Add arc customization: new Helm collector

sophsoph321:
the following arc customizations are in this PR
- "arc mode" which is the clusterType property in the yaml file
- New collector class that runs HELM commands


builder/Dockerfile
```diff
@@ -6,8 +6,13 @@ RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor github.com/Azure/aks-periscope
 
 FROM alpine
 RUN apk --no-cache add ca-certificates
+RUN apk add curl openssl bash --no-cache
 ADD https://storage.googleapis.com/kubernetes-release/release/v1.16.0/bin/linux/amd64/kubectl /usr/local/bin/kubectl
 RUN chmod +x /usr/local/bin/kubectl
+RUN curl -fsSl -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3
```
Tatsinnit, line 12: 💡 Please explore the possible `featureflag` to only trigger when it is needed for the feature.

cmd/aks-periscope/aks-periscope.go
```diff
@@ -14,6 +15,7 @@ import (
 
 func main() {
 	zipAndExportMode := true
+	//exporter := &exporter.LocalMachineExporter{}
```
Tatsinnit, line 18: 💡Nit: Is this intentional? might be removed.

cmd/aks-periscope/aks-periscope.go
```diff
@@ -22,25 +24,38 @@ func main() {
...
+	log.Printf("Storage Account: %s", storageAccount)
 	collectors := []interfaces.Collector{}
 	containerLogsCollector := collector.NewContainerLogsCollector(exporter)
 	collectors = append(collectors, containerLogsCollector)
-	systemLogsCollector := collector.NewSystemLogsCollector(exporter)
-	collectors = append(collectors, systemLogsCollector)
 	networkOutboundCollector := collector.NewNetworkOutboundCollector(5, exporter)
 	collectors = append(collectors, networkOutboundCollector)
-	ipTablesCollector := collector.NewIPTablesCollector(exporter)
-	collectors = append(collectors, ipTablesCollector)
```
Tatsinnit, line 33: 💡 Not sure but, why is this copy pasted to down below? if not intentional lets aim for only new feature work to be effected.

* Intent of asking this question is to also make sure that we document that in case of `AzureArc` why the `IPtables` logs et. al. will not be required?

cmd/aks-periscope/aks-periscope.go
```diff
@@ -56,16 +71,20 @@ func main() {
...
 		}(c)
 	}
 
 	waitgroup.Wait()
 
 	diagnosers := []interfaces.Diagnoser{}
-	diagnosers = append(diagnosers, diagnoser.NewNetworkConfigDiagnoser(dnsCollector, kubeletCmdCollector, exporter))
 	diagnosers = append(diagnosers, diagnoser.NewNetworkOutboundDiagnoser(networkOutboundCollector, exporter))
 
+	if clusterType != "connectedcluster" {
```
Tatsinnit, line 84: 💡Note:  In line 49: https://github.com/Azure/aks-periscope/pull/51/files#diff-cdb15eaf022fc76984176a66a52183b8c3cb0710176cc9349519c5d43fb1d2ffR49 we check the condition with `equality` and at this line we check with `inequality` both intends to do the same check so why not keep it consistently.

cmd/aks-periscope/aks-periscope.go
```diff
@@ -56,16 +71,20 @@ func main() {
...
 			}
+
 			waitgroup.Done()
 		}(c)
 	}
 
 	waitgroup.Wait()
 
 	diagnosers := []interfaces.Diagnoser{}
-	diagnosers = append(diagnosers, diagnoser.NewNetworkConfigDiagnoser(dnsCollector, kubeletCmdCollector, exporter))
```
Tatsinnit, line 66: 💡 Question for doc reasons: why `Arc` will not require `NetworkConfigDiagnoser`?

integration/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM golang AS builder
```
Tatsinnit, line 1: 💡 is this file moved or completely new? What is it trying to achieve? (I think we have one Dockerfile already in place)

pkg/diagnoser/diagnoser.go
```diff
@@ -6,15 +6,17 @@ import "github.com/Azure/aks-periscope/pkg/interfaces"
 type Type int
 
 const (
+	//ConfigValidator defines ConfigValidator Diagnoser Type
+	ConfigValidator Type = iota
 	// NetworkConfig defines NetworkConfig Diagnoser Type
-	NetworkConfig Type = iota
```
Tatsinnit, line 10: 💡 Wondering why `iota` is removed? I believe it is the `const` use purposes, do you know much in detail that removing this will not cause issues existing feature please?

Tatsinnit (changes requested):
💡 Thank you Sophie, I have added basic very early comments, and for the feature flag choice is your so I like the idea which you have regarding `connectedCluster` 

few things as a checklist to make sure you have covered existing behaviours are:

:white_check_mark: Make sure you locally run CI.
:white_check_mark: Tip to locally test changes via image reside here: https://github.com/Azure/aks-periscope#programming-guide 

:white_check_mark: If you find any new functionality could be easily extracted to UT (unit test) then feel enabled to do so.
:white_check_mark: Once we have done all the above, we should have enough confidence to open PR for review. 

(At least these are my initial thoughts) I will add more in case I forget anything.

Really appreciate it. Other folks will kindly add more. 🙏

deployment/aks-periscope.yaml
```diff
@@ -114,15 +116,15 @@ metadata:
   name: containerlogs-config
   namespace: aks-periscope
 data:
-  DIAGNOSTIC_CONTAINERLOGS_LIST: kube-system
```
Tatsinnit, line 117: 💡 Why is this `kube-system` remove? will this have any impact on existing functionality this tool provide?

pkg/diagnoser/diagnoser.go
sophsoph321: Sorry, this was unintentional. I will add it back.

integration/Dockerfile
sophsoph321: I was adding some unit tests earlier and tried testing it using a different Dockerfile. This new Dockerfile should not be merged to master.

cmd/aks-periscope/aks-periscope.go
sophsoph321: The main reasoning was that the kubeletCmdCollector will be turned off if we enable "arc mode". Since NetworkConfigDiagnoser depends on the kubeletCmdCollector, we would not need NetworkConfigDiagnoser.

cmd/aks-periscope/aks-periscope.go
sophsoph321: Sure. Will change it to be consistent.

cmd/aks-periscope/aks-periscope.go
sophsoph321: This is a new exporter I will eventually be submitting a PR for. I am planning to submit a PR for each new class I add in order to decouple my changes to your codebase.

deployment/aks-periscope.yaml
sophsoph321: I changed it because for arc customizations, we would need to get container logs and kube-objects from the azure-arc namespace. So I removed kube-system as the default namespace and changed it to azure-arc. The same thing applies to kube-object config. Of course, we can have azure-arc and kube-system as both of the default namespaces. This was something I was going to ask during the next sync with my team and your team, how we plan on going about setting the default namespace in the yaml. I believe Safeer was also looking into the possibility of using a Helm Chart for periscope.

cmd/aks-periscope/aks-periscope.go
sophsoph321: I copied and pasted to down below because we would be turning off the systemLogs, IPTables, etc. collectors for the connected cluster type (i.e. when arc mode is enabled), as outlined in our design doc. So yes, this was intentional. I will get back to you on the reasoning as to why we do not intend on reusing these collectors.

deployment/aks-periscope.yaml
```diff
@@ -14,10 +14,10 @@ kind: ClusterRole
 metadata:
   name: aks-periscope-role
 rules:
-- apiGroups: ["","metrics.k8s.io"]
-  resources: ["pods", "nodes"]
+- apiGroups: ["","metrics.k8s.io", "apiextensions.k8s.io"]
+  resources: ["pods", "nodes", "customresourcedefinitions", "secrets"]
```
safeermohammed, line 18: revert these changes

deployment/aks-periscope.yaml
```diff
@@ -69,7 +69,7 @@ spec:
         beta.kubernetes.io/os: linux
       containers:
       - name: aks-periscope
-        image: aksrepos.azurecr.io/staging/aks-periscope:v0.3
+        image: aksrepos.azurecr.io/staging/aks-periscope:v0.2
```
safeermohammed, line 72: >v0.2 [](http://example.com/codeflow?start=57&length=4)

What is this change?

deployment/aks-periscope.yaml
safeermohammed: These values can be overwritten by CLI, we would not need these change. @sophsoph321, can you please revert this change? 

---
In reply to: [626987398](https://github.com/Azure/aks-periscope/pull/51#discussion_r626987398) [](ancestors = 626987398)

deployment/aks-periscope.yaml
```diff
@@ -132,6 +134,14 @@ metadata:
 data:
   DIAGNOSTIC_NODELOGS_LIST: /var/log/azure/cluster-provision.log /var/log/cloud-init.log
 ---
+apiVersion: v1 
+kind: ConfigMap 
+metadata:
+  name: clustertype-config
+  namespace: aks-periscope
+data:
+  CLUSTER_TYPE: connectedcluster
```
safeermohammed, line 143: >connectedcluster [](start = 16, length = 16)

Default this to managedCluster

deployment/aks-periscope.yaml
```diff
@@ -114,15 +116,15 @@ metadata:
...
+  DIAGNOSTIC_CONTAINERLOGS_LIST: azure-arc
 ---
 apiVersion: v1 
 kind: ConfigMap 
 metadata:
   name: kubeobjects-config
   namespace: aks-periscope
 data:
-  DIAGNOSTIC_KUBEOBJECTS_LIST: kube-system/pod kube-system/service kube-system/deployment
+  DIAGNOSTIC_KUBEOBJECTS_LIST: azure-arc/pod azure-arc/service azure-arc/deployment
```
safeermohammed, line 127: >DIAGNOSTIC_KUBEOBJECTS_LIST: azure-arc/pod azure-arc/service azure-arc/deployment [](start = 2, length = 81)

Please revert these changes, these can be overwritten via CLI for Arc

safeermohammed:
> 

Nit: revert this newline change

---
Refers to: deployment/aks-periscope.yaml:173 in 6372851. [](commit_id = 6372851519c6733d5786188e31d6d03c170fd4c4, deletion_comment = False)

builder/Dockerfile
safeermohammed: We may download this in helm_collector collect method itself

---
In reply to: [626964750](https://github.com/Azure/aks-periscope/pull/51#discussion_r626964750) [](ancestors = 626964750)

pkg/diagnoser/diagnoser.go
davidkydd: https://golang.org/ref/spec#Iota looks like as long as the first entry in the const has "= iota" defined then additional entries with "= iota" won't have any effect (within that same const). 

Don't think it really matters in this case, but general best practise is to add new values to the end of the enum so as not to change any of the existing mappings, if adding multiple at once then alphabetical ordering of the _newly added_ entries is probably a good idea (in lieu of any other ordering).

deployment/aks-periscope.yaml
```diff
@@ -76,6 +76,8 @@ spec:
         envFrom:
         - configMapRef:
             name: containerlogs-config
+        - configMapRef:
```
Tatsinnit, line 79: 💡 Just a note and observation to keep in mind: 

* These yaml changes for master will affect few consuming tools, like `vscode` ecosystem and `az-cli`, so I will emphasise to keep this in mind for design/ logic flow.
* Just as thoughts: If this gets any complicated, I would recommend exploring separate `yaml` idea.

Thanks guys.

cmd/aks-periscope/aks-periscope.go
sophsoph321: Update: I reverted these changes in my most recent PR update since David had another PR to expose the feature flag for collectors, diagnosers, and exporters

deployment/aks-periscope.yaml
safeermohammed: Thanks Tatsat, In general we are trying to keep the yaml generic.

Instead of having a copy of yaml, does it makes sense to have a helm chart? CLUSTER_TYPE=appliance is other scenairo that might be interested in periscope (appliance is another type of AKS cluster). When this comes in we would end up with one more yaml file.

I have given a first short at helm chart here

[https://github.com/safeermohammed/aks-periscope/tree/master/charts/azure-k8s-periscope](https://github.com/safeermohammed/aks-periscope/tree/master/charts/azure-k8s-periscope)

This if for reference, let me know if this makes sense?

safeermohammed: Also let us know if vscode breaks when this RP is merged, we can create a copy of arc specific yaml file for now.

davidkydd: Ahhhh! Yes thanks @safeermohammed - helm chart is just what we need here. 

Will pull this into my PR and adapt planned changes to leverage it. We can deprecate the existing aks-periscope.yaml and leave as-is until cli + vscode can move to consume the helm chart instead.

Tatsinnit: 💡 Note: We need to make sure that the impact to the external tool, at-least to `az-cli` and `vscode` eco-system is **non-breaking** in case it **if we** need a discuss why will need those breaking changes at all and what are the alternatives.

davidkydd: @Tatsinnit @safeermohammed with a single deployment yaml, which currently is pulled on demand by vscode and CLI (both of which are fixed to target master HEAD), how would you suggest we go about merging breaking changes to the deployment yaml - such as https://github.com/Azure/aks-periscope/pull/52

deployment/aks-periscope.yaml
```diff
@@ -80,6 +80,8 @@ spec:
             name: kubeobjects-config
         - configMapRef:
             name: nodelogs-config
+        - configMapRef:
+            name: clustertype-config
```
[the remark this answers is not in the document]
Tatsinnit: The current `master` for deployment flag has no new flag. This flag seems like the new flag : `clustertype-config` which is added as part of another refactor PR which is mostly little re-design as well.

There are 2 things which is fairly diverged here:

* If you guys 100% want to wait for refactor PR then wait because it is changing so fast that we need to understand how big as impact and what are the key decisions it needs to step back as.
* If this PR is based in Master (which seems its no longer), but in case it will be then feel free to remove it as this flag is not in master.

Key things which other PR really need to think is:

* Use of Kustomize v Helm - We need to especially keep in mind the `Cons` of helm and the security issues which comes along.
* I think the Other PR needs to be in small chunks.

Seems like there seems like there is no-needed chain of dependency created here from a work which is still in review.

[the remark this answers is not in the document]
sophsoph321: @safeermohammed @Tatsinnit  Thanks for your attention on this. I pushed up some new changes just now, which should revert all of my changes to the yaml file as discussed separately with Safeer. 
The main focus of this PR should now be on the new Helm collector I added. This new Helm collector feature has already been tested locally and with the CI pipeline. My PR also includes a unit test with tests the Helm collector feature. 
Please let me know if there's more changes that need to be made before we open this PR for other to review.

PR #54, opened 2021-05-13
Helm feature

sophsoph321:
This PR includes implementation for the following customizations for azure arc:
- In the yaml file, we introduced the ClusterType property in as one of the configmaps, which takes values such as ConnectedClusters, ManagedClusters, ArcAppliances etc. 
- New Helm Collector class (see paragraph below)

**Description of the new HelmCollector class:** 
This HelmCollector class is needed as one of the arc customizations for periscope because as part of the arc extensions feature, the extensions are deployed as helm packages. Therefore, for troubleshooting purposes, the HelmCollector class will be running the commands "helm list --all-namespaces" and "helm history". The command "helm list --all-namespaces" will list all of the releases across all namespaces that are deployed or failed. The command "helm history" will list the historical revisions of a release and provides information such as the following: revision numbers, date and time of the revision, its status, the name of the release chart, the version of the app, and the description.


cmd/aks-periscope/aks-periscope.go
```diff
@@ -22,25 +23,33 @@ func main() {
...
+	systemLogsCollector := collector.NewSystemLogsCollector(exporter)
+	ipTablesCollector := collector.NewIPTablesCollector(exporter)
+	nodeLogsCollector := collector.NewNodeLogsCollector(exporter)
 	kubeletCmdCollector := collector.NewKubeletCmdCollector(exporter)
-	collectors = append(collectors, kubeletCmdCollector)
 	systemPerfCollector := collector.NewSystemPerfCollector(exporter)
-	collectors = append(collectors, systemPerfCollector)
+	helmCollector := collector.NewHelmCollector(exporter)
+
+	if clusterType == "connectedCluster" {
```
jonathan-innis, line 44: Is it possible that this could come with different casing? Maybe use `strings.EqualsFold`?

tests/helm_test.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/Azure/aks-periscope/pkg/utils"
+)
+
+func TestHelm(t *testing.T) {
+
+	output, err := utils.RunCommandOnContainer("helm", "list", "--all-namespaces")
```
jonathan-innis, line 12: Is there a way/an ask for you to write a test where you create a helm collector and then run `Collect()` on it, this would be the most comprehensive form of testing

jonathan-innis (commented):
Couple small things

pkg/collector/helm_collector.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/Azure/aks-periscope/pkg/utils"
+	"path/filepath"
+)
+
+// HelmCollector defines a Helm Collector struct
+type HelmCollector struct {
+	BaseCollector
+}
+
+var _ interfaces.Collector = &HelmCollector{}
```
jonathan-innis, line 14: What is this throw-away variable for?

sophsoph321: It's to verify that the HelmCollector class implements the Collector interface. This assignment tests if a type implements an interface. It's going to give you compile errors if you don't implement the Collector interface.

Tatsinnit:
💡 Thank you so much for this new PR, I have a feeling that This PR over-rides the old PR: https://github.com/Azure/aks-periscope/pull/51 please feel free to close the Old PR.

Please let me know once this PR is ready for review? Thanks.

sophsoph321:
@Tatsinnit so sorry for not getting back in advance! Yes, I have opened this PR so you and the periscope team can review it, it is no longer a draft. I will close PR #51 since this one has the latest changes. Thanks for your attention.

tests/helm_test.go
davidkydd: from a discussion @sophsoph321 had out of band:
When I tested helmCollector.Collect in an isolated unit, I was getting this error: Fail to run command on host: exec: "nsenter": executable file not found in %PATH%

We should look to make the dependency on nsenter injectable or able to be abstracted away for unit testing purposes

cmd/aks-periscope/aks-periscope.go
```diff
@@ -22,25 +23,33 @@ func main() {
...
-	systemLogsCollector := collector.NewSystemLogsCollector(exporter)
-	collectors = append(collectors, systemLogsCollector)
 	networkOutboundCollector := collector.NewNetworkOutboundCollector(5, exporter)
 	collectors = append(collectors, networkOutboundCollector)
-	ipTablesCollector := collector.NewIPTablesCollector(exporter)
-	collectors = append(collectors, ipTablesCollector)
-	nodeLogsCollector := collector.NewNodeLogsCollector(exporter)
-	collectors = append(collectors, nodeLogsCollector)
 	dnsCollector := collector.NewDNSCollector(exporter)
 	collectors = append(collectors, dnsCollector)
```
davidkydd, line 34: This format means the whole block needs to be read to work out which collectors are enabled for a particular mode (connectedCluster vs default). 

Prefer to see all the "append" calls inside the if statement, even though it means more lines and some duplication, it makes it clearer exactly which collectors are enabled for each mode

davidkydd (approved):
Thanks - looks good!

davidkydd (commented):
Thanks - looks good!

tests/helm_test.go
```diff
@@ -0,0 +1,23 @@
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/Azure/aks-periscope/pkg/utils"
+)
+
+func TestHelm(t *testing.T) {
```
Tatsinnit, line 10: 💡 Thank you for this. What is the unit of work this specific test is trying?

* Currently it seems like this test only calls the `utils.RunCommandOnContainer` for `helm` commands and `error` or `printf`, if you mention the intention of this test we can help in grooming this test to better fit the feature need please?

Also, think of this UT as what's the one thing you would like to test as unit of work via this test? (Which intern covers the current added functionality as well)

deployment/aks-periscope.yaml
```diff
@@ -132,6 +134,14 @@ metadata:
 data:
   DIAGNOSTIC_NODELOGS_LIST: /var/log/azure/cluster-provision.log /var/log/cloud-init.log
 ---
+apiVersion: v1 
+kind: ConfigMap 
+metadata:
+  name: clustertype-config
+  namespace: aks-periscope
+data:
+  CLUSTER_TYPE: managedCluster
```
Tatsinnit, line 143: 💡 This default value of `managedClsuter` is unused i.e.  never existed, so lets keep `defaultValue` keep this as **empty** **because** within your code you only check for the `connectedCluster` there is not comparison of `managedCluster` anywhere in the code, so this used `defaultValue` kind of looses its use. 

Could do like what `secret` flags do:

` CLUSTER_TYPE: # <used as flag arc cluster>`

Tatsinnit (changes requested):
💡 Thank you Sophie,  few things as a checklist to make sure you have covered existing behaviours are:

- [ ] Make sure you locally run CI. 
- [ ] Tip to locally test changes via image reside here: https://github.com/Azure/aks-periscope#programming-guide 

- [ ] If you find any new functionality could be easily extracted to UT (unit test) then feel enabled to do so. (We could make it part of the CI as well) 
- [ ] Once we have done all the above, we should have enough confidence to open PR for review. 

Really appreciate it. 🙏

tests/helm_test.go
sophsoph321: @Tatsinnit, this is a test that's used to test the Helm collector functionality. The purpose of the Helm collector is to run the helm commands "helm list" and "helm history". I have looked into making this test more meaningful by directly calling the helmCollector.Collect() function, but as discussed offline, the helmCollector.Collect() implementation logic relies on some helper functions that call RunCommandOnHost(). RunCommandOnHost() executes the "nsenter" command, and since "nsenter" is a unix command, when I test the Collect() function in an isolated unit I got this error: "Fail to run command on host: exec: "nsenter": executable file not found in %PATH%. 

One of my teammates also raised the same question in a previous, will also reopen the thread

Tatsinnit: 💡 Yes, all good, in case we don't find much co-relation between actual unit of work we could test - I would recommend taking this unit-test out for now. 

Essentially, scaffolding a code which a function will already run is kind duplicating the functionality rather then actual test. So, if we think the UT is not achieving anything but testing the obvious then worth to revisit with better idea. 

* I am going to test this tonight and tomorrow, if there is anything else specific then I will loop back but we can avoid any test which is not actually unit of work test but just a retro-fit over a functionality.

sophsoph321: @Tatsinnit, I know the reasons for those two bugs you are running into.
1. Regarding the first issue with "curl" not being found in $PATH$, this can be fixed by running the following command: **apk add curl openssl bash --no-cache**. Please let me know if this command should be feature flagged. 
2. With the second issue where "secrets is forbidden", I have gotten that error message before and it can be worked around by adding secrets to the resources list in the yaml file. However, I don't know if this would cause security issues or other vulnerabilities if we change the yaml like that, so I took it out. But sure, I will ask the arc team if there's a better way to get around it. 
secrets

cmd/aks-periscope/aks-periscope.go
sophsoph321: @davidkydd, just pushed new changes. If it looks better now, we can resolve this thread.

deployment/aks-periscope.yaml
sophsoph321: @Tatsinnit , just pushed new changes that fixes this. If it looks better now, we can resolve this thread.

cmd/aks-periscope/aks-periscope.go
davidkydd: perfect - thanks!

deployment/aks-periscope.yaml
Tatsinnit: Thanks @sophsoph321 thanks, I have a minor recommendation which you could commit as one of last commits to use much better scaffolding name I was thinking:

* `# <custom flag>` - I am thinking of this as a scaffolding for the other consuming tools which use a `replace` and use it for consumptions purposes.

tests/helm_test.go
Tatsinnit: Cool sounds good for 1, regarding point 2, I am sure there could be more specific list command we could use to get something specific out in `arc` case but as mentioned I will leave you guys decide. Thanks.

pkg/collector/helm_collector.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Collect implements the interface method
+func (collector *HelmCollector) Collect() error {
+	rootPath, err := utils.CreateCollectorDir(collector.GetName())
+	if err != nil {
+		return err
+	}
+
+	output, err := utils.RunCommandOnContainer("apk", "add", "curl", "openssl", "bash", "--no-cache")
```
Tatsinnit, line 33: 🐛 Why is this needed here? `Dockerfile` already have it?

deployment/aks-periscope.yaml
sophsoph321: Sure. I just pushed a change that accounts for this change to use `# <custom flag>` as the clusterType value

tests/helm_test.go
sophsoph321: I remember having a discussion with @safeermohammed offline about this. If I remember correctly, we did agree that "helm list --all-namespaces" is the list command we need to run.

cmd/aks-periscope/aks-periscope.go
```diff
@@ -22,25 +23,38 @@ func main() {
...
+	ipTablesCollector := collector.NewIPTablesCollector(exporter)
+	nodeLogsCollector := collector.NewNodeLogsCollector(exporter)
 	kubeletCmdCollector := collector.NewKubeletCmdCollector(exporter)
-	collectors = append(collectors, kubeletCmdCollector)
 	systemPerfCollector := collector.NewSystemPerfCollector(exporter)
-	collectors = append(collectors, systemPerfCollector)
+	helmCollector := collector.NewHelmCollector(exporter)
+
+	if strings.EqualFold(clusterType, "connectedCluster") {
+		collectors = append(collectors, containerLogsCollector)
```
Tatsinnit, line 41: 💡 This all is duplicated now, you could do it this way:

i.e. for conditional if, you have at least 4 duplicate appends. 

Could do something like this: 

```
	collectors = append(collectors, containerLogsCollector)
	collectors = append(collectors, dnsCollector)
	collectors = append(collectors, kubeObjectsCollector)

	if strings.EqualFold(clusterType, "connectedCluster") {
		collectors = append(collectors, helmCollector)

	} else {
		collectors = append(collectors, systemLogsCollector)
		collectors = append(collectors, ipTablesCollector)
		collectors = append(collectors, nodeLogsCollector)
		collectors = append(collectors, kubeletCmdCollector)
		collectors = append(collectors, systemPerfCollector)
	}
```

davidkydd: @Tatsinnit this was originally as you describe above - I suggested changing it to the current form so that all the appends were inside the conditional if / else (not sure where comment has gone).

I think I prefer this over what you are suggesting - otherwise there is a "common" area which has to be modified in the case that we want to add another value for clusterType. 

Don't have strong views either way if @sophsoph321 wants to change it though :)

pkg/collector/helm_collector.go
Tatsinnit, line 33: 💡 🐛 I plan to add PR to your fork tonight, but see I noticed something interesting: 

`:=` is Short variable declaration, **but** if you see carefully, up until line 53 you don't do anything with the `output` var.

I can make small updates and send small PR to your fort by tonight. Thanks.

pkg/collector/helm_collector.go
```diff
@@ -0,0 +1,86 @@
...
+		return err
+	}
+
+	_, err = utils.RunCommandOnContainer("./get_helm.sh")
+	if err != nil {
+		return err
+	}
+
+	helmListFile := filepath.Join(rootPath, "helm_list")
+	output, err := utils.RunCommandOnContainer("helm", "list", "--all-namespaces")
```
Tatsinnit, line 54: 💡 🐛 There are few things not going to work with this `helm` command, and I will discuss in detail with some suggestions, since its a daemon set, we could achieve similar idea with `kubectl` I will discuss this with right folks before moving forward.

In the mean time, can you test this in any `ARC-Enabled` cluster as I believe logs will be empty every time. There is a`kubectl` command which can get right data for it.

pkg/collector/helm_collector.go
```diff
@@ -0,0 +1,88 @@
...
+		return err
+	}
+
+	output, err = utils.RunCommandOnContainer("./get_helm.sh")
+	if err != nil {
+		return err
+	}
+
+	helmListFile := filepath.Join(rootPath, "helm_list")
+	output, err = utils.RunCommandOnContainer("helm", "list", "--all-namespaces")
```
Tatsinnit, line 54: 🐛 More details here: https://github.com/Azure/aks-periscope/pull/54#discussion_r634838600

Tatsinnit: 💡 Let's just dig a little deeper for this: I think adding `secrets` to the list might be ok, and non-impacting, as this all runs with users permission and user in use will run it and that might enable the `Arc` - `helm` scenarios as well.

Tatsinnit: Lets add the `secrets` in resources and keep 1 yaml, I am not able to find out any issues with it, inc are we did we can always have a fallback idea anyways. Thank you. here is the small PR for it: https://github.com/sophsoph321/aks-periscope/pull/1 (Mainly to simplify all this)

davidkydd: This looks like a *very* risky set of permissions to me? This creates a ClusterRole with list permissions for secrets across all namespaces in the cluster. 

What is this needed for? Is there any way we can scope the permissions down to e.g. enable secret get to just that secret which is required?

e.g. number 1 risky permission here: https://www.cyberark.com/resources/threat-research-blog/securing-kubernetes-clusters-by-eliminating-risky-permissions
https://darkbit.io/blog/the-power-of-kubernetes-rbac-list

sophsoph321: @davidkydd  The reasoning comes from here: https://helm.sh/docs/topics/rbac/#example-grant-a-user-read-only-access-at-the-cluster-scope 
If you go all the way at the bottom of that link above, it says in order to run "helm list --all-namespaces", the user needs cluster-scope read access, which requires having both view and secret-reader access. I discussed this internally with arc team, and we couldn't figure out a different way to work around the issues Tats was having with running "helm list".

Tatsinnit: Yeah, +1 to this, essentially this needs to be explored, it feels like helm commands are restrictive this way, but Worth a discussion. (Wider arc discussion)

I would prefer this to be pure kubectl driven: 

* kubectl -n azure-arc get deployments,pods 

Above command will give much better, but if he’ll command I 100% needed we need to make sure we cover these small things and rationales it well.

safeermohammed: For kubectl driven we already have KubeObject to get deployments, pods. This is specific with Helm as it is capturing helm history and its revisions. We discussed about cluster scope access specific to Arc Scenarios earlier (I think this was in terms of CRD access).  Two options that were discussed:

1. Update Periscope clusterRole to include additional permissions
https://github.com/Azure/aks-periscope/blob/master/deployment/aks-periscope.yaml#L12

2. Instead of updating clusterRole (so as to stick to be least privilege for AKS), The CLI will update the rolebinding after periscope deployment to provide additional access to service account

I would have preferred option 1, provided we take consent from user. We can use option 2 for time being if 1 is no go.

Tatsinnit: Yeah, I was thinking if we are keeping this way then we can keep current PR level implementation as it is and with the privacy policy along with case that user need to run this tool.

https://github.com/Azure/aks-periscope#data-privacy-and-collection

safeermohammed: Sure, we will have data privacy link

safeermohammed: @davidkydd, Can you share some more context on security concern on more permission with periscope? Is it a generic comment or some issue which periscope encountered in past. Since periscope gets deleted once the log collection is done, wanted to understand the security concerns in detail. 

Like I mentioned above we have two choices, given by default in clusterrole def or given CLI after installing periscope. I do not see much difference between two as  permission would be need to be given in either case by user on cluster to run the helm list cmd (Probably for CRDs as well in future, once we reach to this collector). More details on this will help on the choice on how and by which component to enable these role permissions.   

My assumption today is periscope is already letting user know what is going to do on the cluster with consent, hence asking the addition role permission concerns in detail.

davidkydd: not due to something experienced in the past , just jumped out as a red flag. Also note that while CLI deployment of periscope will be cleaned up once complete, non-CLI driven deployments aren't cleaned up automatically, and would potentially leave this dangling clusterRole.

Assumed you would be able to at least scope the secrets you need to read down so there wouldn't be need for full "list" permission on a clusterRole, but from the link @sophsoph321 posted it doesn't look like it. If this is already at least-privilege possible for the commands the helm collector needs to run then think thats probably fine.

As we will likely encounter with the CRD permissions - it seems we are forced to either:

1) introduce additional complexity (Arc CLI has to update roleBinding after deploy), or 
2) violate least privilege by giving all flavours of deployments the full super-set of required permissions (AKS deployments of periscope will end up with CRD + secrets list permissions when it makes no use of them)

Again, a problem which could be avoided by using a different config / deployment / yaml file for each integration

builder/Dockerfile
```diff
@@ -6,8 +6,12 @@ RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor github.com/Azure/aks-periscope
 
 FROM alpine
 RUN apk --no-cache add ca-certificates
+RUN apk add --no-cache curl openssl bash
 ADD https://storage.googleapis.com/kubernetes-release/release/v1.16.0/bin/linux/amd64/kubectl /usr/local/bin/kubectl
 RUN chmod +x /usr/local/bin/kubectl
+RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 \
```
davidkydd, line 12: this step adds about 50MB to the final periscope image size (100MB originally). Ideally we would have separate images for helm vs. non helm but this is simpler. Potentially revisit later

pkg/collector/helm_collector.go
Tatsinnit: Yeah, initial thought process to be careful was exactly what I thought, but essentially the use case for this tool will not have any security impact to the behaviour.

User will always run this tool on their known cluster context. I would see it as Helm enablement requirement since this issue common with `Helm` command. 

To get more idea/opinion, I reached out to @JunSun17 (**Thank you** for reply) - as long as we are not logging any secrets in an exported file. `I think the tool can access secrets and use it in the program to perform some logic, just make sure the secret values are not written out when exporting, or it will be exposed.` 

Further, @safeermohammed kindly agrees to stay involved and if there is any need be `Arc` will own and we can together make move forward.

Side note: I would say, we can go ahead given we have captured all the opinion and have plan-b in case we find any concrete scenario. Also, if this at all is any security bug in `Helm` then it will be `helm` issue and in that case I would rather prefer to disable the whole `helm` command.

Thanks guys. 🙏

builder/Dockerfile
Tatsinnit: Sounds good.

pkg/collector/helm_collector.go
davidkydd: Thanks @Tatsinnit - my concern was mostly driven by my own naivety re: k8s security :) 

Believe this can be resolved if nothing else from Arc side

JunSun17: @Tatsinnit thanks for summarizing this! I totally agree with you on the security side.

davidkydd: As a quick note regarding "just make sure the secret values are not written out when exporting, or it will be exposed." Secrets *will* be exported out, but only those corresponding to helm releases, which is by design

builder/Dockerfile
```diff
@@ -6,8 +6,12 @@ RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor github.com/Azure/aks-periscope
 
 FROM alpine
 RUN apk --no-cache add ca-certificates
+RUN apk add --no-cache curl openssl bash
```
arnaud-tincelin, line 9: could we merge lines 8 and 9?

pkg/collector/helm_collector.go
```diff
@@ -0,0 +1,61 @@
...
+
+	err = utils.WriteToFile(helmListFile, output)
+	if err != nil {
+		return err
+	}
+
+	collector.AddToCollectorFiles(helmListFile)
+
+	helmHistoryFile := filepath.Join(rootPath, "helm_history")
+	output, err = utils.RunCommandOnContainer("helm", "history", "-n", "default", "azure-arc")
```
arnaud-tincelin, line 48: Rather than adding some packages on the Docker image and executing command lines from our code, should we use libraries instead ? 
https://github.com/helm/helm/blob/0af54d0f5c0f2a108506733b64752039ee39d165/cmd/helm/chart_list.go#L33
https://github.com/helm/helm/blob/0af54d0f5c0f2a108506733b64752039ee39d165/cmd/helm/history.go#L54

That way we can have better error handling and smaller docker image.

Last point: it is best practice to create new variables rather than doing variables reuse (`output, err =` )

arnaud-tincelin (changes requested):
There are 18 commits, do we squash them in one before merging or do we rewrite a concise history?

builder/Dockerfile
Tatsinnit: 💡 thanks, sounds kuch better, If we are going to use helm package we will remove docker file addition: https://github.com/Azure/aks-periscope/pull/54#discussion_r636775650

Or shall we make this part of our refactor, although helm being new addition worth using the helm package.

pkg/collector/helm_collector.go
sophsoph321: Thank you @arnaud-tincelin, for taking the time to review my changes, these are very good suggestions. 
Also tagging relevant people: @davidkydd @safeermohammed @Tatsinnit 

Regarding the suggestion about utilizing helm library, but one thing to keep in mind is that we, the arc team, are trying to get an MVP out by end of May (a.k.a. next week) which will include this new helm feature for periscope. Refactoring the helm classes so that it uses library functions seems to be a big change and the changes that I have so far have already been tested extensively.

After discussing offline with the periscope team, we had a suggestion for periscope team to make the collectors use library functions even for the kubectl commands in the code. So the helm collector can be taking care of as part of this refactoring.  

For now, given the short amount of time we have left until the MVP needs to be released, my thinking is to freeze the code changes for the time being. But of course, I will do my best to fix the other smaller changes such as creating the new variables and combining the lines in the Dockerfile. 

Please let me know your thoughts on this. 

Of course,

davidkydd: Later refactoring to utilize helm + kubectl from libs rather than invoking executables sounds good to me

Tatsinnit: +1, yes we should do it as part of refactor, I will add a workitem with more details by eod.

I will test any new changes sometime Monday our time or earlier. (Will kee you guys posted)

thanks guys 🙏

Tatsinnit:
> There are 18 commits, do we squash them in one before merging or do we rewrite a concise history?

Thank you for noticing this @arnaud-tincelin , @sophsoph321 if you notice any old item in issue log, it seems you have a commit which references very old issue in this repo issue log. (Which seems like incorrect) 

it will be worth to make commits at-least not referencing the issue which are not related. For example go to the issue log and open any old item and you will see your commit referencing tha workitem. 

like:
D40EF544-ED51-45E5-9916-497E39141AD1

Thanks, 🙏

sophsoph321:
> > There are 18 commits, do we squash them in one before merging or do we rewrite a concise history?
> 
> Thank you for noticing this @arnaud-tincelin , @sophsoph321 if you notice any old item in issue log, it seems you have a commit which references every old issue in this repo issue log. (Which is incorrect)
> 
> For example go to the issue log and open any old item and you will see your commit referencing that workitem. Let’s fix this.

> Thanks, 🙏

@Tatsinnit @arnaud-tincelin. Thanks for bringing this up. I squashed a lot of my commits, so this PR should now have 2 commits instead of 18. 

About that commit which references all of the old issues, it comes from a different branch in my repo, so it won't affect this PR. It might have happened when I was trying to squash some commits in a different branch, I'm relatively new to the squash functionality for git commits and at one point, the commits doubled, which was not supposed to happen. Anyways, I'm trying to change the commit message right now so that it doesn't reference every issue, but am really struggling to get my updates in. Do you have any tips on how to fix this? Thanks.

Tatsinnit:
> > > There are 18 commits, do we squash them in one before merging or do we rewrite a concise history?
> > 
> > 
> > Thank you for noticing this @arnaud-tincelin , @sophsoph321 if you notice any old item in issue log, it seems you have a commit which references every old issue in this repo issue log. (Which is incorrect)
> > For example go to the issue log and open any old item and you will see your commit referencing that workitem. Let’s fix this.
> 
> > Thanks, 🙏
> 
> @Tatsinnit @arnaud-tincelin. Thanks for bringing this up. I squashed a lot of my commits, so this PR should now have 2 commits instead of 18.
> 
> About that commit which references all of the old issues, it comes from a different branch in my repo, so it won't affect this PR. It might have happened when I was trying to squash some commits in a different branch, I'm relatively new to the squash functionality for git commits and at one point, the commits doubled, which was not supposed to happen. Anyways, I'm trying to change the commit message right now so that it doesn't reference every issue, but am really struggling to get my updates in. Do you have any tips on how to fix this? Thanks.

Thanks @sophsoph321 , ah, I think this commit which is appearing in all the issues is from this branch: https://github.com/sophsoph321/aks-periscope/commits/helm_class 

So not from your current branch, I would say if this “helm_class” is not in need you can clean or you can rebase that branch for whatever purpose it was created. 

75886949-DFDE-42C4-A7F4-15AA3381A3C8

this is the specific commit in that branch:

1D00574A-CF40-42E8-874F-C3D3B371AC7D

sophsoph321:
> > > > There are 18 commits, do we squash them in one before merging or do we rewrite a concise history?
> > > 
> > > 
> > > Thank you for noticing this @arnaud-tincelin , @sophsoph321 if you notice any old item in issue log, it seems you have a commit which references every old issue in this repo issue log. (Which is incorrect)
> > > For example go to the issue log and open any old item and you will see your commit referencing that workitem. Let’s fix this.
> > 
> > 
> > > Thanks, 🙏
> > 
> > 
> > @Tatsinnit @arnaud-tincelin. Thanks for bringing this up. I squashed a lot of my commits, so this PR should now have 2 commits instead of 18.
> > About that commit which references all of the old issues, it comes from a different branch in my repo, so it won't affect this PR. It might have happened when I was trying to squash some commits in a different branch, I'm relatively new to the squash functionality for git commits and at one point, the commits doubled, which was not supposed to happen. Anyways, I'm trying to change the commit message right now so that it doesn't reference every issue, but am really struggling to get my updates in. Do you have any tips on how to fix this? Thanks.
> 
> Thanks @sophsoph321 , ah, I think this commit which is appearing in all the issues is from this branch: https://github.com/sophsoph321/aks-periscope/commits/helm_class
> 
> So not from your current branch, I would say if this “helm_class” is not in need you can clean or you can rebase that branch for whatever purpose it was created.

Thanks @Tatsinnit. I have removed the helm_class branch from my repo since it is not needed. Unfortunately, the commit reference is still there, with a warning message saying that "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."

Tatsinnit:
All good @sophsoph321 let's treat this as separate issue. I think the issue is that whatever your other branch has, it is not cleaned-up properly, hence even though that commit does't show up now, the rouge commit stays in that branch in your fork. Thanks.

builder/Dockerfile
Tatsinnit: 💡 @sophsoph321 lets make this change and then please double check the checklist, and we should plan to merge this.

One-liner will become : (Line 8, 9)

* `RUN apk --no-cache add ca-certificates curl openssl bash`

deployment/aks-periscope.yaml
```diff
@@ -132,6 +134,14 @@ metadata:
 data:
   DIAGNOSTIC_NODELOGS_LIST: /var/log/azure/cluster-provision.log /var/log/cloud-init.log
 ---
+apiVersion: v1 
+kind: ConfigMap 
+metadata:
+  name: clustertype-config
+  namespace: aks-periscope
+data:
+  CLUSTER_TYPE: "managedCluster" # <custom flag>
```
SanyaKochhar, line 143: A request: would it be possible to use a generic name here like `FEATURE_FLAGS`? Specifically because in PR #48, based on feedback from @Tatsinnit, we are looking to reuse your flag for adding in 2 new collectors (SMI and OSM) that are not based on cluster type.

Ref: https://github.com/Azure/aks-periscope/pull/48/files#diff-cdb15eaf022fc76984176a66a52183b8c3cb0710176cc9349519c5d43fb1d2ffR49-R58

builder/Dockerfile
sophsoph321: @Tatsinnit,  I have fixed this change in my most recent push. I have reviewed the checklist and have tested my changes locally along with running the CI

deployment/aks-periscope.yaml
Tatsinnit: 💡 Let's do this change in your PR, I think `ARC` folks have almost completed this, so as it looks like only thing you need is renaming this flag and we can do that as part of the `OSM x SMI` workitem.

Tatsinnit (approved):
Thank you @sophsoph321 , these changes looks good, I am approving it and we need to cut a release notes etc, before we formally release it.

Next step for me will be cut a release notes, plan a release. Also, we might have to add small read-me to the main doc as well, but I could do that and get your thoughts.

We need to do `squash merge` so I will do that tonight one @arnaud-tincelin or @JunSun17 has taken on final eyeballing for this PR.

Thanks. 🙏

sophsoph321:
> Thank you @sophsoph321 , these changes looks good, I am approving it and I need to cut a release notes etc, before we formally release it.
> 
> Next step for me will be cut a release notes, plan a release. Also, we might have to add small read-me to the main doc as well, but I could do that and get your thoughts.
> 
> We need to do `squash merge` so I will do that tonight one @arnaud-tincelin or @JunSun17 has taken on final eyeballing for this PR.
> 
> Thanks. 🙏

@Tatsinnit. This sounds good. About the documentation, the most compelling I can think of for now is updating the Appendix, so that the users know that they can customize values for the new configmap ClusterType. Would this new clusterType feature flag impact the az aks kollect cli? If yes, then then there should be an additional example in the User guide section of the ReadMe would also need to be updated. I will let you know if I think of other updates that need to be made for the documentation.

Tatsinnit:
> @Tatsinnit. This sounds good. About the documentation, the most compelling I can think of for now is updating the Appendix, so that the users know that they can customize values for the new configmap ClusterType. Would this new clusterType feature flag impact the az aks kollect cli? If yes, then then there should be an additional example in the User guide section of the ReadMe would also need to be updated. I will let you know if I think of other updates that need to be made for the documentation.

Thank you @sohpsoph321 , I think that sounds like good idea, a doc with the name of `azure-arc.md` what is enabled and what it achieves sounds like a good reading guide under the `docs` folder. 

Your work should not effect current functionality, as long as you have tested your changes agains vanilla deployment, because the new flag is only used to trigger `arc` mode on.  

Idealistically we can add documentation as part of the release cut we will do. Thank you 🙏

deployment/aks-periscope.yaml
SanyaKochhar: Sounds good to me. We can make that change as a fast follow once this gets merged

cc: @shalier

JunSun17 (approved):

PR #57, opened 2021-05-19
Local Machine exporter

sophsoph321:
This is another arc customization that we plan to include, which is to export logs to the local machine in the scenario where there is no network connectivity to Azure.

### DIFF
--- a/cmd/aks-periscope/aks-periscope.go
+++ b/cmd/aks-periscope/aks-periscope.go
@@ -13,9 +13,13 @@ import (
 	"github.com/Azure/aks-periscope/pkg/utils"
 )
 
+const (
+	connectedCluster = "connectedCluster"
+)
+
 func main() {
 	zipAndExportMode := true
-	exporter := &exporter.AzureBlobExporter{}
+	exporters := []interfaces.Exporter{}
 
 	err := utils.CreateCRD()
 	if err != nil {
@@ -23,6 +27,17 @@ func main() {
 	}
 
 	clusterType := os.Getenv("CLUSTER_TYPE")
+	isConnectedCluster := strings.EqualFold(clusterType, connectedCluster)
+	storageAccountName := os.Getenv("AZURE_BLOB_ACCOUNT_NAME")
+	sasTokenName := os.Getenv("AZURE_BLOB_SAS_KEY")
+
+	if isConnectedCluster && (len(storageAccountName) == 0 || len(sasTokenName) == 0) {
+		exporters = append(exporters, &exporter.LocalMachineExporter{})
+	} else {
+		exporters = append(exporters, &exporter.AzureBlobExporter{})
+	}
+
+	exporter := exporters[0]
 
 	collectors := []interfaces.Collector{}
 	containerLogsCollector := collector.NewContainerLogsCollector(exporter)
@@ -36,7 +51,7 @@ func main() {
 	systemPerfCollector := collector.NewSystemPerfCollector(exporter)
 	helmCollector := collector.NewHelmCollector(exporter)
 
-	if strings.EqualFold(clusterType, "connectedCluster") {
+	if isConnectedCluster {
 		collectors = append(collectors, containerLogsCollector)
 		collectors = append(collectors, dnsCollector)
 		collectors = append(collectors, helmCollector)
--- a/pkg/exporter/localmachine_exporter.go
+++ b/pkg/exporter/localmachine_exporter.go
@@ -1,0 +1,71 @@
+package exporter
+
+import (
+	"archive/zip"
+	"io"
+	"log"
+	"os"
+
+	"github.com/Azure/aks-periscope/pkg/interfaces"
+	"github.com/Azure/aks-periscope/pkg/utils"
+)
+
+// LocalMachineExporter defines an Local Machine Exporter
+type LocalMachineExporter struct{}
+
+var _ interfaces.Exporter = &LocalMachineExporter{}
+
+// Export implements the interface method
+func (exporter *LocalMachineExporter) Export(files []string) error {
+	hostName, err := utils.GetHostName()
+	if err != nil {
+		return err
+	}
+
+	outFile, err := os.Create(hostName + ".zip")
+	if err != nil {
+		return err
+	}
+
+	defer outFile.Close()
+
+	zipWriter := zip.NewWriter(outFile)
+	defer zipWriter.Close()
+
+	for _, file := range files {
+		log.Printf("File %s is being exported to local machine", file)
+		fileToZip, err := os.Open(file)
+		if err != nil {
+			return err
+		}
+		defer fileToZip.Close()
+
+		// Get the file information
+		info, err := fileToZip.Stat()
+		if err != nil {
+			return err
+		}
+
+		header, err := zip.FileInfoHeader(info)
+		if err != nil {
+			return err
+		}
+
+		// Using FileInfoHeader() above only uses the basename of the file. If we want
+		// to preserve the folder structure we can overwrite this with the full path.
+		header.Name = file
+
+		// Change to deflate to gain better compression
+		header.Method = zip.Deflate
+
+		writer, err := zipWriter.CreateHeader(header)
+		if err != nil {
+			return err
+		}
+		_, err = io.Copy(writer, fileToZip)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}


cmd/aks-periscope/aks-periscope.go
```diff
@@ -14,83 +15,91 @@ import (
 
 func main() {
 	zipAndExportMode := true
-	exporter := &exporter.AzureBlobExporter{}
+	exporters := []interfaces.Exporter{}
+	clusterType := os.Getenv("CLUSTER_TYPE")
+
+	exporters = append(exporters, &exporter.AzureBlobExporter{})
```
jonathan-innis, line 21: If this is always appended just init wiht it already there, i.e. `[]interfaces.Exporter{&exporter.AzureBlobExporter{}}`

cmd/aks-periscope/aks-periscope.go
```diff
@@ -14,83 +15,91 @@ import (
...
+	if strings.EqualFold(clusterType, "connectedCluster") {
```
jonathan-innis, line 22: Make `connectedCluster` a Const

pkg/exporter/localmachine_exporter.go
```diff
@@ -0,0 +1,66 @@
...
+	zipWriter := zip.NewWriter(outFile)
+	defer zipWriter.Close()
+
+	for _, file := range files {
+		log.Printf("File %s is being exported to local machine", file)
+		fileToZip, err := os.Open(file)
+		if err != nil {
+			return err
+		}
+		defer fileToZip.Close()
```
jonathan-innis, line 38: Just a heads up, this `defer` will not be called until the end of the `func` so you will be opening a lot of file descriptors in the `range` function until you close them at the end. This could potentially be an issue with a large number of files

jonathan-innis: This may be consistent with how others are doing it, in that case not an issue 😄

jonathan-innis (commented):
Small comments, overall looks good 😄

arnaud-tincelin:
> This is another arc customization that we plan to include, which is to export logs to the local machine in the scenario where there is no internet connectivity.

Could you elaborate more on this subject please? Blob Storage can still be accessed by private endpoints even if you block outbound to internet

sophsoph321:
> > This is another arc customization that we plan to include, which is to export logs to the local machine in the scenario where there is no internet connectivity.
> 
> Could you elaborate more on this subject please? Blob Storage can still be accessed by private endpoints even if you block outbound to internet

> > This is another arc customization that we plan to include, which is to export logs to the local machine in the scenario where there is no internet connectivity.
> 
> Could you elaborate more on this subject please? Blob Storage can still be accessed by private endpoints even if you block outbound to internet

@arnaud-tincelin. Good catch, I will update the description. This is an arc customization, a new exporter that we want to have that can download logs to the local machine. Of course, on the periscope end, the logs would be stored into the pods/containers. We're planning on having an arc-specific CLI that's going to use "kubectl cp" to copy the logs from the pods to the local machine. 

About the Azure blob storage, I agree that Azure blob storage will still work for arc connected clusters. But there are arc scenarios where we're also dealing with resources that do not have network connectivity with Azure. When network connectivity to Azure is not possible, we wouldn't be able to validate or access the storage accounts, so the logs would need to be exported to the local machine instead, which is why we're adding this new exporter. Does this make sense?

pkg/exporter/localmachine_exporter.go
sophsoph321: Yes, I might leave this part as is for now, bc other classes seem to be using defer like this. If it does become an issue, I will fix it.

arnaud-tincelin:
> > > This is another arc customization that we plan to include, which is to export logs to the local machine in the scenario where there is no internet connectivity.
> > 
> > 
> > Could you elaborate more on this subject please? Blob Storage can still be accessed by private endpoints even if you block outbound to internet
> 
> > > This is another arc customization that we plan to include, which is to export logs to the local machine in the scenario where there is no internet connectivity.
> > 
> > 
> > Could you elaborate more on this subject please? Blob Storage can still be accessed by private endpoints even if you block outbound to internet
> 
> @arnaud-tincelin. Good catch, I will update the description. This is an arc customization, a new exporter that we want to have that can download logs to the local machine. Of course, on the periscope end, the logs would be stored into the pods/containers. We're planning on having an arc-specific CLI that's going to use "kubectl cp" to copy the logs from the pods to the local machine.
> 
> About the Azure blob storage, I agree that Azure blob storage will still work for arc connected clusters. But there are arc scenarios where we're also dealing with resources that do not have network connectivity with Azure. When network connectivity to Azure is not possible, we wouldn't be able to validate or access the storage accounts, so the logs would need to be exported to the local machine instead, which is why we're adding this new exporter. Does this make sense?

Thank you for your reply @sophsoph321 :) When you speak about local machine, is it the node? If so, do not forget they may be deleted at any time, and the logs gone. We might want to store them on a PV instead. Also, if there is no connectivity, how do you plan to export the logs?

deployment/aks-periscope.yaml
```diff
@@ -76,6 +76,8 @@ spec:
         envFrom:
         - configMapRef:
             name: containerlogs-config
+        - configMapRef:
+            name: clustertype-config
         - configMapRef:
```
safeermohammed, line 81: Take latest from master branch. This changes are already in there

safeermohammed:
Hi @arnaud-tincelin, local machine here is the machine on which az CLI runs. Basically the zip file should be copied to the CLI machine. Both storage and local zip are valid scenarios for Arc and default is always to storage account. Local machine is special case. Offline I will send you specs for Arc.

@sophsoph321, The higher order of priority for zip file (both Azureblob export or local export) is zip files should not stored by default on Nodes. Nodes in Arc are owned by customer and we would not have access. Can you point me to the code where we are making sure in Arc mode the zips are not stored on Nodes?

---
In reply to: [847815210](https://github.com/Azure/aks-periscope/pull/57#issuecomment-847815210) [](http://example.com/codeflow?ancestors=847815210,847815210)

cmd/aks-periscope/aks-periscope.go
```diff
@@ -14,83 +15,91 @@ import (
...
+		exporters = append(exporters, &exporter.LocalMachineExporter{})
+	}
+
 	var waitgroup sync.WaitGroup
 
 	err := utils.CreateCRD()
 	if err != nil {
 		log.Printf("Failed to create CRD: %+v", err)
 	}
+	for _, exporter := range exporters {
```
safeermohammed, line 32: What is this change?

safeermohammed (changes requested):
:clock1:

cmd/aks-periscope/aks-periscope.go
sophsoph321: Ok will fix this then.

arnaud-tincelin: Actually I think it could be fine as the user could specify multiples exporters.
However, I don't understand why for each exporter we run the collectors and the diagnosers? To me, it should be run collectors --> run diagnosers --> run exporters

pkg/exporter/localmachine_exporter.go
```diff
@@ -0,0 +1,66 @@
...
+var _ interfaces.Exporter = &LocalMachineExporter{}
+
+// Export implements the interface method
+func (exporter *LocalMachineExporter) Export(files []string) error {
+	hostName, err := utils.GetHostName()
+	if err != nil {
+		return err
+	}
+
+	outFile, err := os.Create(hostName + ".zip")
```
arnaud-tincelin, line 25: It would be nice if we did something with this error

pkg/exporter/localmachine_exporter.go
```diff
@@ -0,0 +1,66 @@
...
+
+		// Change to deflate to gain better compression
+		header.Method = zip.Deflate
+
+		writer, err := zipWriter.CreateHeader(header)
+		if err != nil {
+			return err
+		}
+		_, err = io.Copy(writer, fileToZip)
+		return err
```
arnaud-tincelin, line 63: The loop will run only once because of this return

cmd/aks-periscope/aks-periscope.go
arnaud-tincelin: I think this is ok if they user wants to use multiple exporters.
However, collectors and diagnosers should run once once for all exporter
the loop should only do something like
```go
for _, exporter := range exporters {
  exporter.Export()
}
```

pkg/exporter/localmachine_exporter.go
arnaud-tincelin, line 25: It would be nice to do something with this error

arnaud-tincelin (changes requested):

arnaud-tincelin:
> Hi @arnaud-tincelin, local machine here is the machine on which az CLI runs. Basically the zip file should be copied to the CLI machine. Both storage and local zip are valid scenarios for Arc and default is always to storage account. Local machine is special case. Offline I will send you specs for Arc.
> 
> @sophsoph321, The higher order of priority for zip file (both Azureblob export or local export) is zip files should not stored by default on Nodes. Nodes in Arc are owned by customer and we would not have access. Can you point me to the code where we are making sure in Arc mode the zips are not stored on Nodes?
> 
> In reply to: [847815210](https://github.com/Azure/aks-periscope/pull/57#issuecomment-847815210)

Keep in mind aks-periscope doesn't not aim to be used as a cli. Exporters shall not be bound as to how we deploy aks-periscope.
I'm available if you want to discuss the use case and solutions we have.

To address your question about zip files, this PR should resolve the issue : https://github.com/Azure/aks-periscope/pull/65

pkg/exporter/localmachine_exporter.go
```diff
@@ -0,0 +1,71 @@
+package exporter
```
Tatsinnit, line 1: 💡Thank you for this, here are my thoughts, I second Arnaud's thoughts here:  https://github.com/Azure/aks-periscope/pull/57#issuecomment-856746917 ; what is this local exporter achieving? 

Is it saving it into the machine the tool is run? (The consuming tool can do that like az cli can have some implementation at their end)  but regarding this tool If that is the case I don't think it will be that easy, the moment we are saving things into end-user's machine we have security and compliance issues, along with clean-up?

safeermohammed: Fair points, From periscope perspective, like Sophie mentioned the zip files are storage in container so that they can be pulled by clients like CLI. The folder where the Clients will Pull the zip files is client local problem. BTW, default is Azure storage, User has to explicitly specify the option if they wish to export the files locally, so the user it aware where and what files are exported on their machine.   If there is a security issue, let us know or point to compliance team so we can close on this aspect.

Regarding scenario, think of case where Pods has misconfigured enterprise proxy setting which is pretty common in Arc case. Azure blobs etc., will not be reachable in this case. Looping @shashankbarsin (Arc PM) on scenarios and questions.

Tatsinnit:
💡 Thanks for this PR, given that #75  is closed now, can you please rebase and merge it with master so that this feature change is ready for review with existing updates. 

Also, please make sure most of the comments are taken care off and let us know and we can add our input.

* Currently, please make sure you add a n description as to what flag will enable this feature? (Might user case scenario is best way$
* What file system level decision this feature will make for the user? I.e. every how writes into Inmemory and transition that to storage. I believe you will write are enabling the feature to be saved in “host machine” where user run this? 

Thank you so much for this and please let us know once this PR is ready for review. 🙏

Tatsinnit:
💡 Gentle ping: @sophsoph321  - is this PR still valid? if it is stale can we please close this.

sophsoph321:
> 💡 Gentle ping: @sophsoph321 - is this PR still valid? if it is stale can we please close this.

Hi @Tatsinnit really sorry for the delayed response. I will eventually refactor this code, but it probably won't be anytime soon. I can close this PR temporarily and will reopen once I've gotten around to making the requested changes. Thanks for your attention on this.

PR #58, opened 2021-05-21
Deploy with Kustomize

arnaud-tincelin:
use Kustomize for deployments

### DIFF
--- a/deployment/aks-periscope.yaml
+++ b/deployment/aks-periscope.yaml
@@ -1,3 +1,5 @@
+# Note: this file is deprecated and will be removed in a future release
+# Use Kustomize to deploy the project
 apiVersion: v1
 kind: Namespace
 metadata:
--- a/docs/kustomize.md
+++ b/docs/kustomize.md
@@ -1,0 +1,50 @@
+# Deploy with Kustomize
+
+To store the logs an Azure Blob Service account is required.
+
+Patch the DeamonSet to add the `AZURE_BLOB_ACCOUNT_NAME` env var:
+
+```yaml
+patches:
+- target:
+    group: apps
+    kind: DaemonSet
+    name: aks-periscope
+    version: v1
+  patch: |-
+    - op: add
+      path: '/spec/template/spec/containers/0/env/-'
+      value:
+        name: AZURE_BLOB_ACCOUNT_NAME
+        value: your_account_name
+```
+
+## Connect to the Storage Account using a SAS key
+
+Create the following secret to connect to the Storage Account using a SAS Key:
+
+```yaml
+secretGenerator:
+- name: azureblob-secret
+  literals:
+  - AZURE_BLOB_SAS_KEY=your_sas_key_base_64_encoded
+
+patches:
+- target:
+    group: apps
+    kind: DaemonSet
+    name: aks-periscope
+    version: v1
+  patch: |-
+    - op: add
+      path: '/spec/template/spec/containers/0/envFrom/-'
+      value: |
+        secretRef:
+          name: azureblob-secret
+```
+
+## Apply
+
+```sh
+kubectl apply -f <(kustomize build)
+```
--- a/deployment/cluster-role-binding.yaml
+++ b/deployment/cluster-role-binding.yaml
@@ -1,0 +1,23 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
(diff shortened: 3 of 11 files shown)


Tatsinnit (changes requested):
💡 Thank you for taking a start on this, fyi: (Just a key observation I will do a depth review latter)

* We need to keep `aks-periscope.yaml` file for separate PR once all this is understood and merged because it is consumed by various external tools like `az-cli` and `vscode` consume this `.yaml`.

docs/appendix.md
```diff
@@ -1,30 +0,0 @@
-# Appendix
```
Tatsinnit, line 1: 💡Idea share: Let's do all the docs update in separate PR, there is history to various docs, and we can cater them all separately.

arnaud-tincelin: I only left the new file docs/kustomize.md and reverted the others

deployment/aks-periscope.yaml
```diff
@@ -83,7 +85,7 @@ spec:
         - secretRef:
             name: azureblob-secret
         volumeMounts:
-        - mountPath: /aks-periscope
+        - mountPath: /log/aks-periscope
```
Tatsinnit, line 88: 💡 minor: do we need to document this anywhere?

deployment/deamon-set.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: apps/v1
```
davidkydd, line 1: typo in file name - daemon vs deamon

deployment/aks-periscope.yaml
davidkydd: @arnaud-tincelin what is this change for? I recommend not changing this file at all due to production downstream dependencies

arnaud-tincelin: What dependencies are your referring to?

Another PR about dockerfile (which must be done first) puts the binary at /aks-periscope so we can't mount a volume there anymore.

Anyway, the best option might be first to remove all dependencies to files in the code so that we don't need this volume at all.

davidkydd: The az cli and vs code extensions both currently pull and deploy the live raw version of this file, and might make other assumptions about its structure:

https://github.com/Azure/vscode-aks-tools/blob/430e8931a06ca6fe59bc8bff9bb98954106237f8/src/commands/periscope/helpers/periscopehelper.ts#L118

https://github.com/Azure/azure-cli-extensions/blob/b613a52b445ee49509803c29622b4eb664d84501/src/aks-preview/azext_aks_preview/custom.py#L1991

I assume this design is to allow the versions of periscope which are deployed by the CLI / vs code to be updated via changes to the periscope repo only - however I don't particularly agree with it and consider it a break waiting to happen, as well as unnecessary friction and complexity  when trying to change periscope itself - would be happier to have dedicated files like /azcli/aks-periscope.yaml and /vscode/aks-periscope.yaml if they do reside in this repo.

Not sure if any of your planned changes would actually break the integrations, but I suggest treating this as a "golden file" and not touching it for the time being - perhaps the kustomize changes could reside in a deployment/kustomize or deployment/dev subdir for the time being?

arnaud-tincelin: az cli & vscode are independent from aks-periscope so we are not going to add anything specific for any tool.
What we could do is to git tag the version of the aks-periscope.yaml file before this PR and the PR on dockerfile so that it matches with a docker image tag.
The only change for azcli & vscode would be in the URL where you would no longer target /latest/ (which is a good thing).
This way we can make any changes without fear to break what is currently in use. What do you think?

docs/kustomize.md
```diff
@@ -0,0 +1,50 @@
...
+- target:
+    group: apps
+    kind: DaemonSet
+    name: aks-periscope
+    version: v1
+  patch: |-
+    - op: add
+      path: '/spec/template/spec/containers/0/env/-'
+      value:
+        name: AZURE_BLOB_ACCOUNT_NAME
```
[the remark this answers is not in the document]
arnaud-tincelin: I have added an example file in /examples

Tatsinnit (approved):
👍 Cool, looks good. Thank you for reverting the `aks-periscope.yaml`  file as it is and as `iteration-1` these kustomize changes look good and not impacting any existing functionality.

We can update few things in iteration-2.

PR #59, opened 2021-05-26
Dockerfile cleanup & fix deployment

arnaud-tincelin:
- cleanup dockerfile & use `ENTRYPOINT`
- fix mount paths

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -1,17 +1,33 @@
-FROM golang AS builder
-RUN mkdir /app
-ADD . /app
-WORKDIR /app
-RUN CGO_ENABLED=0 GOOS=linux go build github.com/Azure/aks-periscope/cmd/aks-periscope
+# Builder
+FROM golang:alpine AS builder
 
+ENV GO111MODULE=on \
+    CGO_ENABLED=0 \
+    GOOS=linux \
+    GOARCH=amd64
+
+WORKDIR /build
+
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+
+COPY . .
+
+RUN go build ./cmd/aks-periscope
+
+# Runner
 FROM alpine
+
 RUN apk --no-cache add ca-certificates curl openssl bash
+
 ADD https://storage.googleapis.com/kubernetes-release/release/v1.16.0/bin/linux/amd64/kubectl /usr/local/bin/kubectl
 RUN chmod +x /usr/local/bin/kubectl
+
 RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 \
     && chmod +x get_helm.sh \
     && ./get_helm.sh
-RUN mkdir /app
-WORKDIR /app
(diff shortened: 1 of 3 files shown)


Tatsinnit:
👍 Looks good, thanks.

I will give it a test from this fork tomorrow, also fyi: @JunSun17 - for name changes and if he can see any other impact.

Tatsinnit (approved):
💡 tested with these changes, with following image: `tatsat/docker-changes`

deployment/aks-periscope.yaml
```diff
@@ -92,13 +92,20 @@ spec:
             name: collectors-config
         - secretRef:
             name: azureblob-secret
+        volumeMounts:
```
Tatsinnit, line 95: 📓 Side note: this needs to plan separately with the release somehow, because of the effect it has to the current feature set.

Tatsinnit (changes requested):
💡 THanks for this, just added one small comment as a note as to what we learned few mins back :)

Tatsinnit (approved):
💡 looks good, I will give a quick local test and merge it in few hours. Thank you for this.

Tatsinnit (changes requested):
💡 Update: Something went wrong and I am still investigating but here is the image created of this branch: `tatsat/docker-changes` - it constantly cause failure, and logs are empty so something before that went wrong.

Screen Shot 2021-07-02 at 10 46 12 AM

builder/Dockerfile
```diff
@@ -1,17 +1,33 @@
...
+
 RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 \
     && chmod +x get_helm.sh \
     && ./get_helm.sh
-RUN mkdir /app
-WORKDIR /app
-COPY --from=builder /app/aks-periscope .
-CMD ["./aks-periscope"]
+
+COPY --from=builder /build/aks-periscope /
```
Tatsinnit, line 31: 🐛 I think `.` lets the bash to identify and run, or might be the `entry point` below needs a mechanism to tell `/` to run with bash. 

Working image with below change: `tatsat/docker-changes-test`

**Fix**

```
COPY --from=builder /build/aks-periscope .

ENTRYPOINT ["./aks-periscope"]
```


Screen Shot 2021-07-02 at 12 01 09 PM

builder/Dockerfile
```diff
@@ -1,17 +1,33 @@
...
+    GOOS=linux \
+    GOARCH=amd64
+
+WORKDIR /build
+
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+
+COPY . .
```
Tatsinnit, line 15: 💡 **quick question**: Just checking for this copy the source and destination is same?

Tatsinnit (approved):
👍 Ok, make sense this is directly associated with the new `kustomisation` approach and we need to involve `az-cli` and `vscode` before we do major release. 

If not, we always have option of `.`. 📓 

Thanks @arnaud-tincelin for contribution.

PR #60, opened 2021-05-26
Ensure the program can stop

arnaud-tincelin:
- use `defer` to ensure the release of work groups
- removed the blocking `select` so the program can exit


Tatsinnit (approved):
💡 Looks good.

Referring to @JunSun17 for another set of eyes. Thanks.

PR #61, opened 2021-05-26
Use zip library to create zip file

arnaud-tincelin:


Tatsinnit (approved):
👍 Thanks for this, I have not tested this yet, but changes look good, especially moving away from running `zip` shell command and using golang lib.

- [ ] I will test it out, but setting it for approval.

Will wait if anyone else wants to drop their thoughts. Thank you 🙏

Tatsinnit (changes requested):
💡 🐛  I found a bug which will be cause because the `*.zip` file is no longer there which was getting consumed by external tools. 

I reckon again, this is the lesson learned from how deep some feature of this tool is engraved.

essentially the `vscode` tool was consuming the zip file to download as direct download feature. 

I am scoping as to how we can get it back.

Tatsinnit:
Thanks @arnaud-tincelin , closing this PR as this https://github.com/Azure/aks-periscope/pull/65 seems to cover all. Please let me know if we need this PR back. 🙏

PR #62, opened 2021-05-31
Add go linter for workflow and repo.

Tatsinnit:
This PR enables the `golang linter` all the changes for this PR is for the existing code and the errors it had. 

Following PR achieves:

* Enabling GoLang Linter in repo.
* Add it to the git-workflow.

Thanks

Old error vs new run looks like this:

**With Changes:**

Screen Shot 2021-05-31 at 6 06 05 PM

**Without changes:**

Screen Shot 2021-05-31 at 4 56 40 PM


arnaud-tincelin:
did you have a look at https://github.com/golangci/golangci-lint-action#how-to-use ?

pkg/diagnoser/networkoutbound_diagnoser.go
```diff
@@ -51,25 +54,26 @@ func (diagnoser *NetworkOutboundDiagnoser) Diagnose() error {
...
 		}
+		defer t.Close()
 
 		dataPoint := networkOutboundDiagnosticDatum{HostName: hostName}
 		scanner := bufio.NewScanner(t)
 		for scanner.Scan() {
 			var outboundDatum collector.NetworkOutboundDatum
-			json.Unmarshal([]byte(scanner.Text()), &outboundDatum)
+			err := json.Unmarshal([]byte(scanner.Text()), &outboundDatum)
+			fmt.Println(err)
```
arnaud-tincelin, line 76: I think we use log.Printf instead of fmt
Also if we are having an error, may be we should have a `continue` to move to the next item ?

pkg/diagnoser/networkoutbound_diagnoser.go
```diff
@@ -51,25 +55,29 @@ func (diagnoser *NetworkOutboundDiagnoser) Diagnose() error {
...
+		defer t.Close()
 
 		dataPoint := networkOutboundDiagnosticDatum{HostName: hostName}
 		scanner := bufio.NewScanner(t)
 		for scanner.Scan() {
 			var outboundDatum collector.NetworkOutboundDatum
-			json.Unmarshal([]byte(scanner.Text()), &outboundDatum)
+			err := json.Unmarshal([]byte(scanner.Text()), &outboundDatum)
+			if err != nil {
+				log.Printf("Unmarshal failed: %+v", err)
```
arnaud-tincelin, line 78: use %w instead :)

pkg/utils/helper.go
```diff
@@ -206,7 +206,10 @@ func CreateCRD() error {
 
 	crdName := "aks-periscope-diagnostic" + "-" + hostName
 
-	writeDiagnosticCRD(crdName)
+	err = writeDiagnosticCRD(crdName)
```
arnaud-tincelin, line 209: small improvement: 
```
if err = writeDiagnosticCRD(crdName); err != nil {
return err
}
```

arnaud-tincelin (approved):

PR #63, opened 2021-06-01
fix azure blob exporter

arnaud-tincelin:
Fix 404 error on azure blob exporter


Tatsinnit (approved):
💡 Looks good! 👍 

Thanks, fyi: @JunSun17 for another set of eyes.

pkg/exporter/azureblob_exporter.go
```diff
@@ -39,28 +42,30 @@ func (exporter *AzureBlobExporter) Export(files []string) error {
 
 	ctx := context.Background()
 
-	pipeline := azblob.NewPipeline(azblob.NewAnonymousCredential(), azblob.PipelineOptions{})
-	accountName := os.Getenv("AZURE_BLOB_ACCOUNT_NAME")
-	sasKey := os.Getenv("AZURE_BLOB_SAS_KEY")
+	cred, err := azblob.NewSharedKeyCredential(accountName, sasKey)
```
Tatsinnit, line 45: 💡 **Question**: will this work with `sasKeyy` seems like it is used with `accountKey` passed,? 

* doc'd here: https://pkg.go.dev/github.com/Azure/azure-storage-blob-go/azblob 

* https://github.com/Azure/azure-storage-blob-go/blob/v0.13.0/azblob/zc_credential_shared_key.go#L21 

Screen Shot 2021-06-03 at 5 27 23 PM

arnaud-tincelin: Indeed, it will not work with a SAS key, I will close the PR

PR #64, opened 2021-06-02
Add support for applications on Azure Stack Cloud

shwetamurali:
aks-periscope is not currently supported on Azure Stack. This change is needed in order to support the aks kollect and kanalyze commands on Azure Stack, which will deploy aks-periscope on the customer cluster. The changes made, such as verifying if the application is running on Azure Stack Cloud, and if so, gathering the cert and storage endpoint suffix and modifying the url accordingly, allows these commands to run on Stack and therefore enables the forwarding of specified node logs to the storage account.


cmd/aks-periscope/aks-periscope.go
```diff
@@ -25,6 +25,13 @@ func main() {
 
 	clusterType := os.Getenv("CLUSTER_TYPE")
 
+	if utils.IsAzureStackCloud() {
+		err := utils.CopyCert()
```
arnaud-tincelin, line 29: Should it rather be something like so that if we can't copy cert we exit? 
```go
if err := utils.CopyCert(); err != nil {
  log.Fataf("Failed to copy cert for Azure Stack Cloud environment: %v", err)
}
```

pkg/utils/helper.go
```diff
@@ -11,6 +12,64 @@ import (
...
+	}
+	cloud := azure.Cloud
+	return strings.ToUpper(cloud) == "AZURESTACKCLOUD"
+}
+
+// CopyCert saves the cert content to a new file inside the container
+func CopyCert() error {
+	cert, err := RunCommandOnHost("cat", "/etc/ssl/certs/azsCertificate.pem")
+	if err != nil {
+		return fmt.Errorf("Failed to retrieve cert content: %+v", err)
```
arnaud-tincelin, line 47: Errors can be concatenate to each others, for instance the final prompt could be in the format `Error: error1: error2: ...` so starting errors message with a lower case is to me a better option + we know we are in an error, so I tend to avoid words such as "error" or "failed".
Finally, there is now error wrapping with `%w` so we should use it instead of `%v`
Example:
```go
if err != nil {
    // Return an error which unwraps to err.
    return fmt.Errorf("decompress %v: %w", name, err)
}
```

pkg/utils/helper.go
```diff
@@ -11,6 +12,64 @@ import (
...
+	}
+	err = WriteToFile("/etc/ssl/certs/azsCertificate.pem", cert)
+	if err != nil {
```
arnaud-tincelin, line 50: you could have a more concise code with 
`if err = WriteToFile("/etc/ssl/certs/azsCertificate.pem", cert); err != nil {`

pkg/utils/helper.go
```diff
@@ -11,6 +12,64 @@ import (
...
+
+// GetStorageEndpointSuffix returns the SES url from the JSON file as a string
+func GetStorageEndpointSuffix() string {
+	if IsAzureStackCloud() {
+		ascFile, err := RunCommandOnHost("cat", "/etc/kubernetes/azurestackcloud.json")
+		if err != nil {
+			return PublicAzureStorageEndpointSuffix
+		}
+		var azurestackcloud AzureStackCloud
+		err = json.Unmarshal([]byte(ascFile), &azurestackcloud)
```
arnaud-tincelin, line 64: same here, you can add the Unmarshal to the if statement

pkg/utils/helper.go
```diff
@@ -11,6 +12,64 @@ import (
 	"strings"
 )
 
+// PublicAzureStorageEndpointSuffix defines default Storage Endpoint Suffix
+const PublicAzureStorageEndpointSuffix = "core.windows.net"
```
arnaud-tincelin, line 16: That would work fine but I am wondering if the user should specify the whole hostname (rather than just the name today) so that we don't have to determine it by ourselves. What do you think?

pkg/utils/helper.go
```diff
@@ -11,6 +12,64 @@ import (
...
+	err = json.Unmarshal([]byte(azureFile), &azure)
+	if err != nil {
+		return false
+	}
+	cloud := azure.Cloud
+	return strings.ToUpper(cloud) == "AZURESTACKCLOUD"
+}
+
+// CopyCert saves the cert content to a new file inside the container
+func CopyCert() error {
```
arnaud-tincelin, line 44: Why do we need to do this?

arnaud-tincelin (changes requested):

cmd/aks-periscope/aks-periscope.go
```diff
@@ -25,6 +25,13 @@ func main() {
 
 	clusterType := os.Getenv("CLUSTER_TYPE")
 
+	if utils.IsAzureStackCloud() {
```
Tatsinnit, line 28: 💡 Minor: Do you think it will be beneficial to move these 3/4 new line of code into separate function and all you need to add is more appropriate name for that funct: like `copyCertIfAzureStackCloud` or something more apt **Or** add couple of line of comments please for clarity.

pkg/utils/helper.go
```diff
@@ -11,6 +12,64 @@ import (
...
+	if err != nil {
+		return false
+	}
+	var azure Azure
+	err = json.Unmarshal([]byte(azureFile), &azure)
+	if err != nil {
+		return false
+	}
+	cloud := azure.Cloud
+	return strings.ToUpper(cloud) == "AZURESTACKCLOUD"
```
Tatsinnit, line 40: 💡 is this `AZURESTACKCLOUD` well known constant, can you please add a doc which has this reference as comment above please, (For case of reference)

pkg/utils/helper.go
```diff
@@ -11,6 +12,64 @@ import (
...
+}
+
+// CopyCert saves the cert content to a new file inside the container
```
jadarsie, line 43: you can probably generalize this function to `CopyFileFromHost` if the file paths are passed as parameters

pkg/utils/helper.go
jadarsie, line 40: `strings.EqualFold` is usually the preferred way of performing case-insensitive comparisons

pkg/utils/helper.go
shwetamurali: If we don't, we get this error: 

`Collector: systemperf, export data failed: Failed to create blob: -> github.com/Azure/azure-pipeline-go/pipeline.NewError, C:/Users/t-smurali/Desktop/aks-periscope/vendor/github.com/Azure/azure-pipeline-go/pipeline/error.go:154
HTTP request failed
Put "https://shweta.blob.redmond.ext-n31r1205.masd.stbtest.microsoft.com/test-cluster1-redmond-cloudapp-ext-n31r1205-masd-stbtest-micros?restype=container&se=2021-08-07T07%3A22%3A25Z&sig=JlwqovhJdorwmnFudse": x509: certificate signed by unknown authority`

It is because we need the self signed cert in order to communicate with the storage account.

pkg/utils/helper.go
shwetamurali: I don't believe there is any documentation on this, "AzureStackCloud" is just the string that appears in the json file under that key if the application is running in Azure Stack. I will make it a string contact and use that reference instead and add a comment.

jadarsie: https://kubernetes-sigs.github.io/cloud-provider-azure/install/configs/#azure-stack-configuration

pkg/utils/helper.go
shwetamurali: @arnaud-tincelin I'm a bit confused by what you mean - how would the user pass it in? Would it be through the yaml file or some other method?

pkg/utils/helper.go
shwetamurali: thanks @jadarsie for the documentation -- looks like `AzureStackCloud `is a well known constant.

pkg/utils/helper.go
arnaud-tincelin: Yes indeed as of today the deployment requires the following secret:
```yaml
---
apiVersion: v1
kind: Secret
metadata:
  name: azureblob-secret
  namespace: aks-periscope
type: Opaque
data:
  AZURE_BLOB_ACCOUNT_NAME: YXRpcGVyaXNjb3Bl
```
we could change it to `AZURE_BLOB_ACCOUNT_HOSTNAME` or `AZURE_BLOB_ACCOUNT_BASE_URL`

PS: this is a simple question, I'm not requesting changes :)

pkg/utils/helper.go
arnaud-tincelin: Great! May I suggest to explain briefly this in the code so that we know why azure stack requires this?

Tatsinnit (approved):

PR #66, opened 2021-06-09
Hack to prevent the program from exiting

arnaud-tincelin:


Tatsinnit (approved):
💡 I see this existed since day 1 of this code : https://github.com/Azure/aks-periscope/blob/b98d66a238e942158ef2628a9315b58937ff9c8f/cmd/aks-periscope/aks-periscope.go#L70

Good to catch this and add it back with comments for future reference and understanding.

Thanks.

PR #69, opened 2021-06-20
WIP - support for KubernetesInDocker (kind) clusters

davidkydd:
test image available @ dakyddacr.azurecr.io/periscope/kind:v11

Currently containerlogs, nodelogs and systemperf failing with:
2021/06/20 10:17:20 Collector: containerlogs, collect data failed: Fail to run command on host: exit status 127
2021/06/20 10:17:20 Collector: nodelogs, collect data failed: Fail to run command on host: exit status 1
2021/06/20 10:17:21 Collector: systemperf, collect data failed: Fail to run command in container: exit status 1: Error from server (Forbidden): unknown (get services http:heapster:)

Other functionality appears present on KIND cluster, no regressions observed on AKS cluster.

These outstanding issues look like ones that @sophsoph321 has already encountered and mentioned


pkg/collector/osm_collector.go
```diff
@@ -17,11 +17,11 @@ type OsmCollector struct {
...
-func NewOsmCollector(exporter interfaces.Exporter) *OsmCollector {
+func NewOsmCollector(exporters []interfaces.Exporter) *OsmCollector {
 	return &OsmCollector{
 		BaseCollector: BaseCollector{
 			collectorType: Osm,
-			exporter:      exporter,
+			exporters:     exporters,
 		},
 	}
 }
```
davidkydd, line 27: interface signature change to an array, no reason it had to be a single exporter

pkg/collector/systemlogs_collector.go
```diff
@@ -15,11 +15,11 @@ type SystemLogsCollector struct {
...
-func NewSystemLogsCollector(exporter interfaces.Exporter) *SystemLogsCollector {
+func NewSystemLogsCollector(exporters []interfaces.Exporter) *SystemLogsCollector {
 	return &SystemLogsCollector{
 		BaseCollector: BaseCollector{
 			collectorType: SystemLogs,
-			exporter:      exporter,
+			exporters:     exporters,
 		},
 	}
 }
```
davidkydd, line 25: interface signature change to an array, no reason it had to be a single exporter

pkg/collector/systemperf_collector.go
```diff
@@ -15,11 +15,11 @@ type SystemPerfCollector struct {
...
-func NewSystemPerfCollector(exporter interfaces.Exporter) *SystemPerfCollector {
+func NewSystemPerfCollector(exporters []interfaces.Exporter) *SystemPerfCollector {
 	return &SystemPerfCollector{
 		BaseCollector: BaseCollector{
 			collectorType: SystemPerf,
-			exporter:      exporter,
+			exporters:     exporters,
 		},
 	}
 }
```
davidkydd, line 25: interface signature change to an array, no reason it had to be a single exporter

pkg/collector/collector.go
```diff
@@ -70,11 +72,15 @@ func (b *BaseCollector) AddToCollectorFiles(file string) {
...
 func (b *BaseCollector) Export() error {
-	if b.exporter != nil {
-		return b.exporter.Export(b.collectorFiles)
+	var result error
+	for _, exporter := range b.exporters {
+		if exporter != nil {
+			if err := exporter.Export(b.collectorFiles); err != nil {
+				result = multierror.Append(result, err)
+			}
+		}
```
davidkydd, line 81: interface signature change to an array, no reason it had to be a single exporter

pkg/exporter/azureblob_exporter.go
```diff
@@ -19,23 +19,60 @@ const (
 )
 
 // AzureBlobExporter defines an Azure Blob Exporter
-type AzureBlobExporter struct{}
+type AzureBlobExporter struct {
+	BaseExporter
```
davidkydd, line 23: make exporter design consistent with collector and diagnoser as there is no reason for it to be different, required in order to add a constructor

pkg/exporter/exporter.go
```diff
@@ -0,0 +1,24 @@
+package exporter
```
davidkydd, line 1: make exporter design consistent with collector and diagnoser as there is no reason for it to be different. This allows new exporters to be defined using the same pattern as collector / diagnose

pkg/utils/helper_test.go
```diff
@@ -0,0 +1,69 @@
...
+    client-certificate: /etc/kubernetes/certs/client.crt
+    client-key: /etc/kubernetes/certs/client.key
+contexts:
+- context:
+    cluster: localcluster
+    user: client
+  name: localclustercontext
+current-context: localclustercontext`,
+"dakydd-test-eastus-dns-d0daedb9.hcp.eastus.azmk8s.io"},
+}
```
davidkydd, line 51: this test case confirms current behaviour is retained

pkg/exporter/azureblob_exporter_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"testing"
+)
+
+var getStorageContainerNameTests = []struct {
+	apiServerFqdn string
+	containerName string
+}{
+	{"dakydd-test-eastus-dns-d0daedb9.hcp.eastus.azmk8s.io", "dakydd-test-eastus-dns-d0daedb9"},
```
davidkydd, line 11: this test confirms that current behaviour of azureblob_exporter remains unchanged

pkg/exporter/exporter.go
```diff
@@ -0,0 +1,24 @@
...
+
+// BaseExporter defines Base Exporter
+type BaseExporter struct {
+	exporterType Type
+}
+
+// GetName gets exporter name
+func (b *BaseExporter) GetName() string {
+	return b.exporterType.name()
+}
```
davidkydd, line 24: make exporter design consistent with collector and diagnoser as there is no reason for it to be different. This allows new exporters to be defined using the same pattern as collector / diagnose

pkg/exporter/azureblob_exporter.go
```diff
@@ -19,23 +19,60 @@ const (
...
+}
```
davidkydd, line 24: make exporter design consistent with collector and diagnoser as there is no reason for it to be different. This allows new exporters to be defined using the same pattern as collector / diagnose

pkg/exporter/azureblob_exporter.go
```diff
@@ -19,23 +19,60 @@ const (
...
-// Export implements the interface method
-func (exporter *AzureBlobExporter) Export(files []string) error {
-	APIServerFQDN, err := utils.GetAPIServerFQDN()
-	if err != nil {
-		return err
+// NewAzureBlobExporter is a constructor
+func NewAzureBlobExporter() *AzureBlobExporter {
+	return &AzureBlobExporter{
+		BaseExporter: BaseExporter{
+			exporterType: AzureBlob,
```
davidkydd, line 32: Constructor required to match collector / diagnoser pattern

pkg/exporter/azureblob_exporter.go
```diff
@@ -19,23 +19,60 @@ const (
...
 	containerName = strings.TrimRight(containerName[:len], "-")
+	return containerName, nil
+}
+
+// Export implements the interface method
+func (exporter *AzureBlobExporter) Export(files []string) error {
+	APIServerFQDN, err := utils.GetAPIServerFQDN()
+	if err != nil {
+		return fmt.Errorf("Failed to get APIServerFQDN: %+v", err)
+	}
```
davidkydd, line 70: same as existing code (originally line 26-30)

pkg/exporter/azureblob_exporter.go
```diff
@@ -19,23 +19,60 @@ const (
...
+	return APIServerFQDN, nil
+}
 
+func (exporter *AzureBlobExporter) GetNonKINDStorageContainerName(APIServerFQDN string) (string, error) {
 	containerName := strings.Replace(APIServerFQDN, ".", "-", -1)
 	len := strings.Index(containerName, "-hcp-")
 	if len == -1 {
 		len = maxContainerNameLength
 	}
 	containerName = strings.TrimRight(containerName[:len], "-")
```
davidkydd, line 61: original getContainerName implementation ends up here

pkg/exporter/azureblob_exporter.go
```diff
@@ -19,23 +19,60 @@ const (
...
+	var err error
+	if utils.IsKubernetesInDocker() {
+		containerName, err = exporter.GetKubernetesInDockerStorageContainerName(APIServerFQDN)
+	} else {
+		containerName, err = exporter.GetNonKINDStorageContainerName(APIServerFQDN)
+	}
+
+	//TODO run a sanitizer over the final chars in the containerName
+	return containerName, err
+}
```
davidkydd, line 49: new method allows calling different implementation for KIND clusters

cmd/aks-periscope/aks-periscope.go
```diff
@@ -11,19 +11,17 @@ import (
 	"github.com/Azure/aks-periscope/pkg/exporter"
 	"github.com/Azure/aks-periscope/pkg/interfaces"
 	"github.com/Azure/aks-periscope/pkg/utils"
+
+	"github.com/hashicorp/go-multierror"
 )
 
 func main() {
-	zipAndExportMode := true
```
davidkydd, line 17: this is always true - remove it

cmd/aks-periscope/aks-periscope.go
```diff
@@ -11,19 +11,17 @@ import (
...
-	exporter := &exporter.AzureBlobExporter{}
```
davidkydd, line 18: replaced with selectExporters method

cmd/aks-periscope/aks-periscope.go
```diff
@@ -11,19 +11,17 @@ import (
...
 
 	err := utils.CreateCRD()
 	if err != nil {
 		log.Fatalf("Failed to create CRD: %v", err)
 	}
 
-	collectorList := strings.Fields(os.Getenv("COLLECTOR_LIST"))
-
```
davidkydd, line 26: moved into selectCollectors

cmd/aks-periscope/aks-periscope.go
```diff
@@ -32,40 +30,166 @@ func main() {
...
+	}
+
+	// TODO: Hack: for now AKS-Periscope is running as a deamonset so it shall not stop (or the pod will be restarted)
+	// Revert from https://github.com/Azure/aks-periscope/blob/b98d66a238e942158ef2628a9315b58937ff9c8f/cmd/aks-periscope/aks-periscope.go#L70
+	select {}
+
+	// TODO: remove this //nolint comment once the select{} has been removed
+	//nolint:govet
+	return
+}
```
davidkydd, line 62: refactor - no functional change

deployment/aks-periscope-kind.yaml
```diff
@@ -0,0 +1,200 @@
+# Note: this file is deprecated and will be removed in a future release
```
davidkydd, line 1: temp for development

cmd/aks-periscope/aks-periscope.go
```diff
@@ -32,40 +30,166 @@ func main() {
...
 	}
 
+	enabledCollectorNames = strings.Fields(os.Getenv("ENABLED_COLLECTORS"))
+
+	for _, collectorName := range enabledCollectorNames {
+		collectors = append(collectors, allCollectorsByName[collectorName])
+	}
+
+	return collectors
+}
```
davidkydd, line 137: introduce new ENABLED_COLLECTORS as intended replacement for COLLECTOR_LIST. Support COLLECTOR_LIST for back-compat

cmd/aks-periscope/aks-periscope.go
```diff
@@ -113,28 +232,29 @@ func main() {
...
+	for _, e := range exporters {
+		if err := e.Export(filesToExport); err != nil {
+			result = multierror.Append(result, err)
 		}
 	}
+	return result
 }
 
-// zipAndExport zip the results and export
-func zipAndExport(exporter interfaces.Exporter) error {
```
davidkydd, line 129: zipAndExport replaced with two functions: zipOutputDirectory & runExporters

pkg/utils/helper.go
```diff
@@ -90,28 +106,59 @@ func GetHostName() (string, error) {
...
 			fqdn := line[index+len("server: "):]
-			fqdn = strings.Replace(fqdn, "https://", "", -1)
-			fqdn = strings.Replace(fqdn, ":443", "", -1)
-			return fqdn, nil
+			fqdnurl, err := url.Parse(fqdn)
+			if err != nil {
+				return "", fmt.Errorf("Fail to parse url from fqdn: %s", fmt.Sprint(err)+": "+fqdn)
+			}
+
+			host, _, err := net.SplitHostPort(fqdnurl.Host)
```
davidkydd, line 121: these are equivalent to the strings.Replace above but will match any port or http://

pkg/utils/helper.go
```diff
@@ -90,28 +106,59 @@ func GetHostName() (string, error) {
 	return strings.TrimSuffix(string(hostname), "\n"), nil
 }
 
-// GetAPIServerFQDN gets the API Server FQDN from the kubeconfig file
-func GetAPIServerFQDN() (string, error) {
```
davidkydd, line 94: replaced with alternative implementation below

pkg/exporter/azureblob_exporter.go
```diff
@@ -19,23 +19,60 @@ const (
...
+
+func (exporter *AzureBlobExporter) GetKubernetesInDockerStorageContainerName(APIServerFQDN string) (string, error) {
+	return APIServerFQDN, nil
```
sophsoph321, line 52: I tested periscope on the kubeadm cluster, which will also call this function along with the kind cluster. Setting the container like this will give errors such as "the resource contains invalid characters". Could you modify this function to the following: 
```
func (exporter *AzureBlobExporter) GetKubernetesInDockerStorageContainerName(APIServerFQDN string) (string, error) {

        containerName := strings.Replace(APIServerFQDN, ".", "-", -1)

	return containerName, nil
}
```

pkg/exporter/azureblob_exporter.go
sophsoph321: The aks-engine cluster will call the original getContainerName and the variable len will be set to maxContainerNameLength=63. This will cause a string out of bounds exception. So before we search for the index, of "-hcp-" (a.k.a. right after line 56), could we add a check like this: 
```
if strings.EqualFold(clusterType, "connectedCluster") {
		return containerName, nil
	}
```

sophsoph321 (changes requested):
Hi @David, thanks for putting together this PR. I made some comments regarding some string manipulation for the blob container name which are important for fixing exceptions that involve the resource name containing invalid characters. But overall looks good!

pkg/exporter/azureblob_exporter.go
davidkydd: sure - although I think I can solve this without needing to involve the clusterType, just by making it robust to a missing "-hcp-". 

Can you provide an example aks-engine APIServerFQDN and resulting containerName you would expect to be returned? I can then add these as cases for getStorageContainerNameTests and make sure your scenario is covered

pkg/exporter/azureblob_exporter.go
davidkydd: will do - again if you have one handy it would be great if you could provide an example of a kubeadm apiServerFQDN and the containerName you would expect as result, so I can use as test data

pkg/exporter/azureblob_exporter.go
davidkydd: Found an example here: https://blog.jreypo.io/2019/06/13/scaling-operations-with-aks-engine/

Using this test data for now:
{apiServerFQDN: "my-cluster.westeurope.cloudapp.azure.com", containerName: "my-cluster-westeurope-cloudapp-azure-com"}.

Please let me know if this is incorrect or inadequate.

pkg/exporter/azureblob_exporter.go
davidkydd: also - can you confirm that on a kubeadm cluster we will call this function due to the kubeconfig being located in "/etc/kubernetes/kubelet.conf"? And hence the helper function currently named "IsKubernetesInDocker" will return true?

If the answer to the above is yes, it should probably necessitate a rename or minor redesign to this function

pkg/exporter/azureblob_exporter.go
sophsoph321: sure. The apiServerFQDN I got was more generic than the one you found. I got something like this: 
{apiServerFQDN: "10.255.255.5", containerName: "10-255-255-5}

pkg/exporter/azureblob_exporter.go
sophsoph321: This is the example for the kubeadm cluster I'm using to test: 
{apiServerFQDN: "troubleshoot.testing.db839748.eastus.cloudapp.azure.com", containername: "troubleshoot-testing-db839748-eastus-cloudapp-azure-com"}

pkg/exporter/azureblob_exporter.go
davidkydd: thanks - will add this case too. Curious that its just an IP address not a url.

Wouldn't think "10-255-255-5" would be a particularly informative containerName, perhaps there are other fields of the kubelet config which would be better to use in the kubeadm scenario, rather than reading "server: <what-we-are-calling-the-fqdn>".

pkg/exporter/azureblob_exporter.go
sophsoph321: And yes, the kubeadm cluster has the kubeconfig located in "/etc/kubernetes/kubelet.conf", so the IsKubernetesInDocker boolean will return true.

pkg/diagnoser/diagnoser.go
```diff
@@ -36,9 +39,13 @@ func (b *BaseDiagnoser) AddToDiagnoserFiles(file string) {
 
 // Export implements the interface method
 func (b *BaseDiagnoser) Export() error {
-	if b.exporter != nil {
-		return b.exporter.Export(b.diagnoserFiles)
+	var result error
+	for _, exporter := range b.exporters {
+		if exporter != nil {
```
arnaud-tincelin, line 44: not sure this if is very usefull
also, I think we could use `errgroup.Group` here and run each exporter in a go func

pkg/exporter/azureblob_exporter.go
```diff
@@ -19,23 +20,68 @@ const (
...
+	containerName = strings.TrimRight(containerName, "-")
+	return containerName, nil
+}
+
 // Export implements the interface method
 func (exporter *AzureBlobExporter) Export(files []string) error {
 	APIServerFQDN, err := utils.GetAPIServerFQDN()
 	if err != nil {
-		return err
+		return fmt.Errorf("Failed to get APIServerFQDN: %+v", err)
```
arnaud-tincelin, line 78: `fmt.Errorf("get APIServerFQDN: %w", err)`

pkg/exporter/azureblob_exporter.go
```diff
@@ -19,23 +20,68 @@ const (
...
 	}
 
-	containerName := strings.Replace(APIServerFQDN, ".", "-", -1)
-	len := strings.Index(containerName, "-hcp-")
-	if len == -1 {
-		len = maxContainerNameLength
+	containerName, err := exporter.GetStorageContainerName(APIServerFQDN)
+	if err != nil {
+		return fmt.Errorf("Failed to get StorageContainerName: %+v", err)
```
arnaud-tincelin, line 83: `return fmt.Errorf("get StorageContainerName: %w", err)`

pkg/exporter/azureblob_exporter.go
```diff
@@ -46,7 +92,7 @@ func (exporter *AzureBlobExporter) Export(files []string) error {
 	ses := utils.GetStorageEndpointSuffix()
 	url, err := url.Parse(fmt.Sprintf("https://%s.blob.%s/%s%s", accountName, ses, containerName, sasKey))
 	if err != nil {
-		return fmt.Errorf("Fail to build blob container url: %+v", err)
+		return fmt.Errorf("Failed to build blob container url: %+v", err)
```
arnaud-tincelin, line 95: `return fmt.Errorf("build blob container url: %w", err)`

pkg/utils/helper.go
```diff
@@ -52,6 +54,20 @@ func IsAzureStackCloud() bool {
 	return strings.EqualFold(cloud, AzureStackCloudName)
 }
 
+// IsKubernetesInDocker returns true if the application is running on KubernetesInDocker (kind)
+func IsKubernetesInDocker() bool {
+	//TODO refactor the conditional logic this check guards into a new "KindClusterOperations" type behind an interface
+	//test the AKS kubeconfig location, if we find something then this isn't a KIND cluster
+	_, err := RunCommandOnHost("ls", "/var/lib/kubelet/kubeconfig")
```
arnaud-tincelin, line 61: they are libraries in golang to do the equivalent as a ls. We should stop using RunCommand

pkg/utils/helper.go
```diff
@@ -90,28 +106,59 @@ func GetHostName() (string, error) {
...
 	for _, line := range lines {
 		index := strings.Index(line, "server: ")
 		if index >= 0 {
 			fqdn := line[index+len("server: "):]
-			fqdn = strings.Replace(fqdn, "https://", "", -1)
-			fqdn = strings.Replace(fqdn, ":443", "", -1)
-			return fqdn, nil
+			fqdnurl, err := url.Parse(fqdn)
+			if err != nil {
+				return "", fmt.Errorf("Fail to parse url from fqdn: %s", fmt.Sprint(err)+": "+fqdn)
```
arnaud-tincelin, line 118: return "", fmt.Errorf("parse url from fqdn %s: %w", fqdn, err)

pkg/utils/helper.go
```diff
@@ -90,28 +106,59 @@ func GetHostName() (string, error) {
...
+			}
+
+			host, _, err := net.SplitHostPort(fqdnurl.Host)
+			if err != nil {
+				return "", fmt.Errorf("Fail to split host port from fqdnurl: %s", fmt.Sprint(err)+": "+fqdnurl.String())
```
arnaud-tincelin, line 123: `return "", fmt.Errorf("split host port from fqd url %s: %w", fqdnurl, err)`

pkg/utils/helper.go
```diff
@@ -90,28 +106,59 @@ func GetHostName() (string, error) {
...
-
 	return "", errors.New("Could not find server definitions in kubeconfig")
 }
 
+//ReadKubeletConfig reads the kubeletConfig from the node
+func ReadKubeletConfig() (string, error) {
+	if IsKubernetesInDocker() {
+		output, err := RunCommandOnHost("cat", "/etc/kubernetes/kubelet.conf")
+		if err != nil {
+			return "", fmt.Errorf("Can't open kubeconfig file at /etc/kubernetes/kubelet.conf\": %+v", err)
```
arnaud-tincelin, line 137: return "", fmt.Errorf("open kubeconfig file at \"/etc/kubernetes/kubelet.conf\": %w", err)

pkg/utils/helper.go
```diff
@@ -90,28 +106,59 @@ func GetHostName() (string, error) {
...
 		}
 	}
-
 	return "", errors.New("Could not find server definitions in kubeconfig")
 }
 
+//ReadKubeletConfig reads the kubeletConfig from the node
+func ReadKubeletConfig() (string, error) {
+	if IsKubernetesInDocker() {
+		output, err := RunCommandOnHost("cat", "/etc/kubernetes/kubelet.conf")
```
arnaud-tincelin, line 135: I think we can avoid runCmd here

pkg/utils/helper.go
```diff
@@ -90,28 +106,59 @@ func GetHostName() (string, error) {
...
+		if err != nil {
+			return "", fmt.Errorf("Can't open kubeconfig file at /etc/kubernetes/kubelet.conf\": %+v", err)
+		}
+		return output, nil
+	} else {
+		output, err := RunCommandOnHost("cat", "/var/lib/kubelet/kubeconfig")
+		if err != nil {
+			return "", fmt.Errorf("Can't open kubeconfig file at /var/lib/kubelet/kubeconfig\": %+v", err)
```
arnaud-tincelin, line 143: `fmt.Errorf("open kubeconfig file at \"/var/lib/kubelet/kubeconfig\": %w", err)`

arnaud-tincelin (changes requested):

pkg/collector/containerlogs_collector_containerd.go
```diff
@@ -0,0 +1,150 @@
...
+			filepath:      path.Join(directoryPath, logFile),
+		})
+	}
+
+	return containerLogs
+}
+
+//GetAllContainerLogFilesThatHaveEverRunOnHost gets the list of log files for all containers that have ever run on the host
+func (collector *ContainerLogsCollectorContainerD) GetAllContainerLogFilesThatHaveEverRunOnHost() ([]string, error){
+	output, err := utils.RunCommandOnHost("ls", containerLogDirectory)
```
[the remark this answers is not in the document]
davidkydd: This is a good idea for commands in the container, do you know how we could do this for RunCommandOnHost (which uses nsenter to run commands in the namespace of the OS hosting the container?)

pkg/diagnoser/diagnoser.go
davidkydd: Yeah not sure it's useful either, just matching the existing shape for now. Can possibly remove this later or change it as you suggest

pkg/utils/helper.go
davidkydd: As above, will this work via nsenter?

pkg/utils/helper.go
davidkydd: As above, how can we do this via nsenter?

davidkydd: Maybe this? https://pkg.go.dev/github.com/kata-containers/runtime/virtcontainers/pkg/nsenter?utm_source=godoc. Regardless this should go in a separate PR

davidkydd: Hmm might not be a go https://www.weave.works/blog/linux-namespaces-and-go-don-t-mix

pkg/exporter/azureblob_exporter.go
davidkydd: Actually wont change this now - it is consistent with the rest of the codebase, if we want to change it should all be at once and separately

sophsoph321 (approved):

cmd/aks-periscope/aks-periscope.go
davidkydd: replaced with ZIP_AND_EXPORT bool in exporters-config

Tatsinnit:
FYI: just a note.

I am not sure why it was merged because bunch of code which was said only for development purpose is not removed? (Like the one below)

F6D4CEEC-A185-4BD1-8DA0-37021921918F


also:

Next time if you are self merging please use: “squash merge” it uses one commit which contains history of all your work and this PR will stay.

advantage:

* it will keep the master commit history cleaner.

thanks,

PR #70, opened 2021-06-20
add default .gitattributes

davidkydd:


Tatsinnit (approved):
💡 Looks good, nothing much to add from my end, wider eyes more discussion here, but +1 from my side. thanks.  

* https://stackoverflow.com/questions/21472971/what-is-the-purpose-of-text-auto-in-gitattributes-file

davidkydd:
I _think_ this change could potentially introduce conflicts (whole file appears to have changed) in helper.go for devs on windows without autocrlf = true. Setting `git config --global core.autocrlf true` should fix the issue in that case, and is generally recommended on windows anyway (see https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration, core.autocrlf section)

Tatsinnit:
> I _think_ this change could potentially introduce conflicts (whole file appears to have changed) in helper.go for those on windows without autocrlf = true. Setting `git config --global core.autocrlf true` should fix the issue in that case, and is generally recommended on windows anyway (see https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration, core.autocrlf section)

Yeah, I did notice that but not sure if it was intentional, sounds good, if intention was to only make change to one file i.e. `.gitignore` and revert other, I can approve and go ahead with merge if this sounds good? Or you area asking for hand for reverting this?

davidkydd:
> > I _think_ this change could potentially introduce conflicts (whole file appears to have changed) in helper.go for those on windows without autocrlf = true. Setting `git config --global core.autocrlf true` should fix the issue in that case, and is generally recommended on windows anyway (see https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration, core.autocrlf section)
> 
> Yeah, I did notice that but not sure if it was intentional, sounds good, if intention was to only make change to one file i.e. `.gitignore` and revert other, I can approve and go ahead with merge if this sounds good? Or you area asking for hand for reverting this?

This should be merged as is, the changes to the other files are only line ending changes, and due to running "git add renormalize" to standardize all text file line endings to LF (what we want the repo side to have, and how it will stay with text=auto gitattribute set). 

Otherwise we risk an inconsistent mish mash of CRLF and LF files repo side, which can lead to erroneous conflicts when we have developers working on both windows and *nix in the same repo.

Issue #67, opened 2021-06-13
feat: add SMI collector in Periscope


johnsonshi:
**Feature request**
Enhance periscope so that it collects SMI information from a cluster.

**Additional context**
This is blocked pending #48 being merged.
Once #48 is merged, I will open a PR that adds SMI collection capability to Periscope. This work has already been implemented in https://github.com/SanyaKochhar/aks-periscope/pull/30

Tatsinnit:
💡🙏 thank you for creating this, FYI regarding #48 : it’s been few days it has merge conflict along with final changes we recommended, if we can fix that we can prioritise the changes and merge. Rest looks good. Thank you.

PR #71, opened 2021-06-22
feat: add SMI log collector

johnsonshi:
Add log collector for Service Mesh Interface
logs, CustomResourceDefinitions, and custom
resources.

Related to PR #48.

Resolves issue #67.

Signed-off-by: Johnson Shi <Johnson.Shi@microsoft.com>


johnsonshi:
Hey @Tatsinnit, I created a draft PR outlining the types of logs of SMI CRDs that I'll be collecting. Let me know what you think plus your thoughts on how the logs are organized.

pkg/collector/smi_collector.go
```diff
@@ -0,0 +1,90 @@
...
+
+	//		Create a directory for the namespace.
+
+	// 		For each Smi CRD:
+
+	// 			Log a TSV list of all custom resources of that Smi CRD type.
+
+	// 			For each custom resource of that Smi CRD type:
+
+	// 				Log the YAML definition of that custom resource of that Smi CRD type.
```
johnsonshi, line 80: @Tatsinnit (and @SanyaKochhar // @shalier) - thoughts on the types of SMI logs I'll collect plus the way I organized the files?

johnsonshi:
Ready now

pkg/collector/smi_collector.go
```diff
@@ -0,0 +1,109 @@
...
+	rootPath, err := utils.CreateCollectorDir(collector.GetName())
+	if err != nil {
+		return err
+	}
+
+	// Filter to obtain a list of Smi CustomResourceDefinitions in the cluster
+	crdNameContainsSmiPredicate := func(s string) bool { return strings.Contains(s, "smi-spec.io") }
+	smiCrdsList := filter(allCrdsList, crdNameContainsSmiPredicate)
+	if len(smiCrdsList) == 0 {
+		// TODO should this return an error?
```
johnsonshi, line 47: Should this return error here?

deployment/aks-periscope.yaml
```diff
@@ -77,7 +80,7 @@ spec:
         beta.kubernetes.io/os: linux
       containers:
       - name: aks-periscope
-        image: aksrepos.azurecr.io/staging/aks-periscope:v0.3
+        image: johshmsft/aks-periscope:latest
```
johnsonshi, line 83: Will reverse this.

deployment/aks-periscope.yaml
```diff
@@ -148,7 +151,7 @@ metadata:
   name: collectors-config
   namespace: aks-periscope
 data:
-  COLLECTOR_LIST: managedCluster # <custom flag>
+  COLLECTOR_LIST: managedCluster, OSM, SMI # <custom flag>
```
johnsonshi, line 154: Should I reverse this?

SanyaKochhar: I think we had decided that by default these collectors will not be run, but can be enabled by users optionally. @Tatsinnit should we be adding documentation around this?

pkg/collector/smi_collector.go
```diff
@@ -0,0 +1,109 @@
...
+		log.Printf("Cluster does not contain any SMI CustomResourceDefinitions\n")
+		return nil
+	}
+
+	collectSmiCrdDefinitions(collector, filepath.Join(rootPath, "smi_crd_definitions"), smiCrdsList)
```
SanyaKochhar, line 52: nit: crdDefinitions = custom resource definition Definitions
```suggestion
	collectSmiCrds(collector, filepath.Join(rootPath, "smi_crd_definitions"), smiCrdsList)
```

johnsonshi: thanks! :)

pkg/collector/smi_collector.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return nil
+}
+
+func collectSmiCrdDefinitions(collector *SmiCollector, rootPath string, smiCrdsList []string) {
+	for _, smiCrd := range smiCrdsList {
+		fileName := smiCrd + "_definition.yaml"
+		kubeCmd := []string{"get", "crd", smiCrd, "-o", "yaml"}
+		if err := collector.CollectKubectlOutputToCollectorFiles(rootPath, fileName, kubeCmd); err != nil {
+			log.Printf("Failed to collect %s: %+v", fileName, err)
```
Tatsinnit, line 63: 💡 is this `fmt.Errorf(` or `log.Printf` I will let you decide, but the message `Failed` suggest its `Error`?

pkg/collector/smi_collector.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+	}
+}
+
+func collectSmiCustomResourcesFromAllNamespaces(collector *SmiCollector, rootPath string, smiCrdsList []string) {
+	// Get all namespaces in the cluster
+	namespacesList, err := utils.GetResourceList([]string{"get", "namespaces", "-o", "jsonpath={..metadata.name}"}, " ")
+	if err != nil {
+		log.Printf("Failed to list namespaces in the cluster: %+v", err)
```
Tatsinnit, line 72: 💡 same as above: Worth identifying that use-case here is `error` or something else?

pkg/collector/smi_collector.go
```diff
@@ -0,0 +1,109 @@
...
+		return
+	}
+
+	for _, namespace := range namespacesList {
+		namespaceRootPath := filepath.Join(rootPath, "namespace_"+namespace)
```
Tatsinnit, line 77: 💡Idea: Worth adding minor comment as to `namespace_*` leads to? Objective with the `join` aka `namespaceRootPath` docs from msdn or something?

pkg/collector/smi_collector.go
```diff
@@ -0,0 +1,109 @@
...
+		collectSmiCustomResourcesFromSingleNamespace(collector, namespaceRootPath, smiCrdsList, namespace)
+	}
+}
+
+func collectSmiCustomResourcesFromSingleNamespace(collector *SmiCollector, rootPath string, smiCrdsList []string, namespace string) {
```
Tatsinnit, line 82: 💡 The name: `collectSmiCustomResourcesFromSingleNamespace` suggestion to me that `SingleNamespace` is key but what is that `singlenamespace` please?

pkg/collector/smi_collector.go
```diff
@@ -0,0 +1,109 @@
...
+	for _, smiCrdType := range smiCrdsList {
+		// get all custom resources of this smi crd type
+		customResourcesList, err := utils.GetResourceList([]string{"get", smiCrdType, "-n", namespace, "-o", "jsonpath={..metadata.name}"}, " ")
+		if err != nil {
+			log.Printf("Failed to list custom resources of type %s in namespace %s: %+v", smiCrdType, namespace, err)
```
Tatsinnit, line 87: 💡 Same as above, decide between `error` or not.

pkg/collector/smi_collector.go
```diff
@@ -0,0 +1,109 @@
...
+			fileName := smiCrdType + "_" + customResourceName + ".yaml"
+			kubeCmd := []string{"get", smiCrdType, customResourceName, "-n", namespace, "-o", "yaml"}
+			if err := collector.CollectKubectlOutputToCollectorFiles(customResourcesRootPath, fileName, kubeCmd); err != nil {
+				log.Printf("Failed to collect %s: %+v", fileName, err)
+			}
+		}
+	}
+}
+
+func filter(ss []string, test func(string) bool) (ret []string) {
```
Tatsinnit, line 102: 💡 `test func(string) bool` --  test? (shall we name this more meaningful and better)? 

* I think motivated from this code here: https://stackoverflow.com/questions/37562873/most-idiomatic-way-to-select-elements-from-an-array-in-golang

pkg/collector/smi_collector.go
```diff
@@ -0,0 +1,109 @@
...
+			log.Printf("Failed to list custom resources of type %s in namespace %s: %+v", smiCrdType, namespace, err)
+			continue
+		}
+
+		customResourcesRootPath := filepath.Join(rootPath, smiCrdType+"_custom_resources")
+		for _, customResourceName := range customResourcesList {
+			fileName := smiCrdType + "_" + customResourceName + ".yaml"
+			kubeCmd := []string{"get", smiCrdType, customResourceName, "-n", namespace, "-o", "yaml"}
+			if err := collector.CollectKubectlOutputToCollectorFiles(customResourcesRootPath, fileName, kubeCmd); err != nil {
+				log.Printf("Failed to collect %s: %+v", fileName, err)
```
Tatsinnit, line 96: 💡 Let's think if this is `error` or not? (Messaging shows error and mostly currently the collector in error define it as error.

Tatsinnit (changes requested):
💡 Thank you so much for this, I have added initial comments below. Really appreciate it.

deployment/aks-periscope.yaml
johnsonshi: Resolved in d34fbb85991d6fa1822bcf29d4b0158c5b3af79d

deployment/aks-periscope.yaml
johnsonshi: Resolved in d34fbb85991d6fa1822bcf29d4b0158c5b3af79d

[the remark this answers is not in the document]
johnsonshi: Resolved in d34fbb85991d6fa1822bcf29d4b0158c5b3af79d

pkg/collector/smi_collector.go
johnsonshi: Resolved in eeb2fdb

pkg/collector/smi_collector.go
johnsonshi: Resolved in 5b56484 - this should be an error. If the user chooses to collect `SMI` data, then the scenario is that the user *has* `SMI` installed in their cluster, which is why they chose to collect `SMI` data. If they explicitly chose `SMI` data through the explicitly setting the flag, then this needs to return error if it fails to find any SMI CRDs in the cluster.

pkg/collector/smi_collector.go
johnsonshi: This is resolved in 5b56484f608330da011cacf14314c75cf620ddad.
This should not return an error and should not give up trying to find the YAML definitions of the other SMI CRDs.
When periscope fails to obtain the YAML definition of a single SMI CRD, it should log a message noting that it skipped collecting the YAML of *only this* SMI CRD (should not error). Periscope should then proceed with collecting the YAML definitions of other SMI CRDs.

```go
	for _, smiCrd := range smiCrdsList {
		fileName := smiCrd + "_definition.yaml"
		kubeCmd := []string{"get", "crd", smiCrd, "-o", "yaml"}
		if err := collector.CollectKubectlOutputToCollectorFiles(rootPath, fileName, kubeCmd); err != nil {
			log.Printf("Skipping: unable to collect yaml definition of %s to %s: %+v", smiCrd, fileName, err)
                         ^^^ Do not give up and error - continue collecting the yaml definitions of the other SMI CRDs.
		}
	}
```

pkg/collector/smi_collector.go
johnsonshi: Resolved in 5b56484.
This should return an error. Periscope needs to list all of the cluster's namespaces in order to proceed with collecting SMI custom resources. If Periscope fails to list the cluster's namespaces (`kubectl get namespaces...`), then something horrible has gone wrong in the cluster, so an error should be returned here.

```
	// Get all namespaces in the cluster
	namespacesList, err := utils.GetResourceList([]string{"get", "namespaces", "-o", "jsonpath={..metadata.name}"}, " ")
	if err != nil {
		return fmt.Errorf("Failed to collect SMI custom resources: unable to list namespaces in the cluster: %+v", err)
	}
```

pkg/collector/smi_collector.go
johnsonshi: Resolved in 5b56484.
Periscope should not give up when it fails to list custom resources of a specific type.
Rather, it should log a message, skip listing custom resources of this type, and move on with listing custom resources of another type in the list.
```go
	for _, smiCrdType := range smiCrdsList {
		// get all custom resources of this smi crd type
		customResourcesList, err := utils.GetResourceList([]string{"get", smiCrdType, "-n", namespace, "-o", "jsonpath={..metadata.name}"}, " ")
		if err != nil {
			log.Printf("Skipping: unable to list custom resources of type %s in namespace %s: %+v", smiCrdType, namespace, err)
			continue
		}
```

pkg/collector/smi_collector.go
johnsonshi: This is resolved in 5b56484.
If periscope is unable to collect the YAML definition of a single custom resource, it should not give up and `error`.
Instead, it should log a message and proceed with collecting the YAML definitions of other custom resources (of the same type).
```go
		customResourcesRootPath := filepath.Join(rootPath, smiCrdType+"_custom_resources")
		for _, customResourceName := range customResourcesList {
			fileName := smiCrdType + "_" + customResourceName + ".yaml"
			kubeCmd := []string{"get", smiCrdType, customResourceName, "-n", namespace, "-o", "yaml"}
			if err := collector.CollectKubectlOutputToCollectorFiles(customResourcesRootPath, fileName, kubeCmd); err != nil {
				log.Printf("Skipping: unable to collect yaml definition of %s (custom resource type: %s) to %s: %+v", customResourceName, smiCrdType, fileName, err)
			}
		}
```

pkg/collector/smi_collector.go
johnsonshi: Resolved in 5075df7.
I renamed the function to `collectSmiCustomResourcesFromNamespace` because the previous name was misleading.
What `collectSmiCustomResourcesFromNamespace` does is collect all smi custom resources from a namespace. The newly-changed function name is more appropriate and less confusing.

pkg/collector/smi_collector.go
johnsonshi: Resolved in e1fb508.
Periscope will iterate through each namespace in the cluster and attempt to collect SMI custom resources from each namespace. If SMI custom resources exist in a namespace, the YAML definitions of each custom resources will be stored in a sub-directory named `namespace_X`.

Here's how the directory structure of SMI collected logs look like.

image

image

image

image

pkg/collector/smi_collector.go
johnsonshi: Resolved with ee5f839

deployment/aks-periscope.yaml
```diff
@@ -28,6 +28,9 @@ rules:
 - apiGroups: ["apiextensions.k8s.io"]
   resources: ["customresourcedefinitions"]
   verbs: ["get", "list", "watch"]
+- apiGroups: ["access.smi-spec.io", "specs.smi-spec.io", "split.smi-spec.io"]
```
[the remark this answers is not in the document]
johnsonshi: Hi @Tatsinnit, thanks for reviewing this, but I do not quite understand your comment? These permissions are required by the Periscope SMI collector in order to:
* obtain the YAML definition of SMI CRDs
* list SMI custom resources of each SMI CRD type
* obtain the YAML definition of each SMI custom resource.

The SMI spec dictates that each service mesh should implement SMI CRDs themselves while conforming to SMI standards. These (below) are examples of SMI CRDs that are installed as part of `Open Service Mesh`. Other service meshes such as `Linkerd` have their own implementation of SMI CRDs, which may be different.

```shell
» kubectl get crds | grep smi                                                                                                                                                    johnsonshi@DESKTOP-AQ60TOB
httproutegroups.specs.smi-spec.io            2021-06-22T21:32:33Z
tcproutes.specs.smi-spec.io                  2021-06-22T21:32:33Z
trafficsplits.split.smi-spec.io              2021-06-22T21:32:33Z
traffictargets.access.smi-spec.io            2021-06-22T21:32:33Z
udproutes.specs.smi-spec.io                  2021-06-22T21:32:33Z
```

Adding additional permissions to enable Periscope to obtain/list/get SMI CRDs shouldn't introduce any new intended behavior. The only positive effect is that this prevents Periscope from running into permissions during SMI log collection.

Regarding testing Periscope's permissions to list/obtain/get SMI CRDs, I think that would be a challenge. The SMI CRDs can only be installed into a cluster by *installing a full service mesh, such as OSM or Linkerd*. Periscope won't be able to run integration tests to test SMI CRD permissions without installing OSM/Linkerd.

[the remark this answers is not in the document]
Tatsinnit: Hiya @johnsonshi not sure about above comment, but all I meant to mention is when you are testing this work with local published image? (Please do test the vanilla scenario? aka without SMI work as well) Thanks.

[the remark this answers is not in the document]
johnsonshi: Oh sure thing! Yeah when I tested it with the locally published image (with SMI flag) it worked. I will try testing with the vanilla periscope image with plus set the corresponding flags and let you know the test results.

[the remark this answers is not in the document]
johnsonshi: @Tatsinnit I've conducted the tests.

As can be seen here, the deployment yaml that I ran had the upstream vanilla aks-periscope image: https://paste.ubuntu.com/p/Mqt2q2v9Wc/

As can be seen here, the deployment yaml that I ran had the additional ClusterRole permissions for smi CRDs. https://paste.ubuntu.com/p/tjjMYPq56R/

This setup (vanilla upstream image WITH the additional SMI permissions in ClusterRole) successfully ran without error: https://paste.ubuntu.com/p/R34G4JF6vR/


----

I have successfully tested my changes with both locally published image and the vanilla upstream image.
Under both scenarios I tested _with_ and _without_ smi flag (`COLLECTOR_LIST: managedCluster OSM SMI # <custom flag>`).
Also under both scenarios I tested _with_ the additional SMI cluster role permissions.

I think this PR is now ready for merging. Let me know if you need anything else.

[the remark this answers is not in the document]
Tatsinnit: Thank you for this, there is branch conflict also I believe ou will notice that there is a simplified collector implementation will help to simplify this more.

Thank you for taking care of SMI, and Thank-you to  @arnaud-tincelin for already sorting out the OSM work for this simplification, please let us know once conflict is resolved.

pkg/collector/smi_collector.go
```diff
@@ -0,0 +1,99 @@
...
+	rootPath, err := utils.CreateCollectorDir(collector.GetName())
+	if err != nil {
+		return err
+	}
+
+	// Filter to obtain a list of Smi CustomResourceDefinitions in the cluster
+	crdNameContainsSmiTestPredicate := func(s string) bool { return strings.Contains(s, "smi-spec.io") }
+	smiCrdsList := utils.FilterSliceElemsWithTestPredicate(allCrdsList, crdNameContainsSmiTestPredicate)
+	if len(smiCrdsList) == 0 {
+		return fmt.Errorf("Cluster does not contain any SMI CRDs")
```
arnaud-tincelin, line 48: I believe this line produces a warning as there is nothing to format. I think you want to use error.New here

pkg/collector/smi_collector.go
```diff
@@ -0,0 +1,99 @@
...
+
+	collectSmiCrds(collector, filepath.Join(rootPath, "smi_crd_definitions"), smiCrdsList)
+	return collectSmiCustomResourcesFromAllNamespaces(collector, filepath.Join(rootPath, "smi_custom_resources"), smiCrdsList)
+}
+
+func collectSmiCrds(collector *SmiCollector, rootPath string, smiCrdsList []string) {
+	for _, smiCrd := range smiCrdsList {
+		fileName := smiCrd + "_definition.yaml"
+		kubeCmd := []string{"get", "crd", smiCrd, "-o", "yaml"}
+		if err := collector.CollectKubectlOutputToCollectorFiles(rootPath, fileName, kubeCmd); err != nil {
```
arnaud-tincelin, line 59: I think you need to rebase as this function does not exist anymore

pkg/collector/smi_collector.go
```diff
@@ -0,0 +1,99 @@
...
+			log.Printf("Skipping: unable to collect yaml definition of %s to %s: %+v", smiCrd, fileName, err)
+		}
+	}
+}
+
+func collectSmiCustomResourcesFromAllNamespaces(collector *SmiCollector, rootPath string, smiCrdsList []string) error {
+	// Get all namespaces in the cluster
+	namespacesList, err := utils.GetResourceList([]string{"get", "namespaces", "-o", "jsonpath={..metadata.name}"}, " ")
+	if err != nil {
+		return fmt.Errorf("Failed to collect SMI custom resources: unable to list namespaces in the cluster: %+v", err)
```
arnaud-tincelin, line 69: This line produces a warning:
- no capital letter
- %w instead of %v
- as you are already in an error message you could simplify the text

`return fmt.Errorf("collect SMI custom resources: unable to list namespaces in the cluster: %w", err)`

arnaud-tincelin (changes requested):

johnsonshi:
Hi @Tatsinnit @arnaud-tincelin, I've resolved all conflicts and addressed @arnaud-tincelin's additional comments. I think this can now be merged - thank you!!! 😀🙏🙏

deployment/aks-periscope.yaml
```diff
@@ -115,8 +118,8 @@ metadata:
   namespace: aks-periscope
 type: Opaque
 data:
-  AZURE_BLOB_ACCOUNT_NAME: # <accountName, base64 encoded>
-  AZURE_BLOB_SAS_KEY: # <saskey, base64 encoded>
+  AZURE_BLOB_ACCOUNT_NAME: YXpzdG9yYWdlYWNjam9oc2g=
+  AZURE_BLOB_SAS_KEY: P3N2PTIwMjAtMDgtMDQmc3M9YmZxdCZzcnQ9c2NvJnNwPXJ3ZGxhY3VwdGZ4JnNlPTIwMjItMDEtMDFUMTc6MTg6MTZaJnN0PTIwMjEtMDctMDhUMDg6MTg6MTZaJnNwcj1odHRwcyZzaWc9bGhMZ0h4Yjl5QWRQZ05XTU1zR2VRZFRnT2F0NXpiJTJCZGgzS3RBbmd1SmE4JTNE
```
Tatsinnit, line 122: ⚠️ Be careful and please rotate this sasKey.

deployment/aks-periscope.yaml
```diff
@@ -77,7 +80,7 @@ spec:
         beta.kubernetes.io/os: linux
       containers:
       - name: aks-periscope
-        image: aksrepos.azurecr.io/staging/aks-periscope:v0.3
+        image: johshmsft/aks-periscope:latest # johshmsft/aks-periscope:latest aksrepos.azurecr.io/staging/aks-periscope:v0.3
```
Tatsinnit, line 83: 💡 I think this needs to be reverted. Since this is your local image.

deployment/aks-periscope.yaml
```diff
@@ -115,8 +118,8 @@ metadata:
   namespace: aks-periscope
 type: Opaque
 data:
-  AZURE_BLOB_ACCOUNT_NAME: # <accountName, base64 encoded>
-  AZURE_BLOB_SAS_KEY: # <saskey, base64 encoded>
+  AZURE_BLOB_ACCOUNT_NAME: YXpzdG9yYWdlYWNjam9oc2g=
```
Tatsinnit, line 121: ⚠️ Due to security reasons I would recommend cycling this account name as well.

tests/aksperiscopeintegration_test.go
```diff
@@ -76,3 +76,14 @@ func checkifpodsrunning(t *testing.T) bool {
 
 	return false
 }
+
+func TestFilterSliceElemsWithTest(t *testing.T) {
```
Tatsinnit, line 80: 💡 please add this into a new test file `helper_test.go` since this function is test of a function in `helper` and not really an integration test.

Tatsinnit (changes requested):
👍 Rest looks good, once you made leftover recommended changes and especially not checking in with credentials :) 

cc: @arnaud-tincelin for rest. Thank you so much, 🙏

pkg/utils/helper.go
```diff
@@ -361,3 +361,14 @@ func writeDiagnosticCRD(crdName string) error {
 
 	return nil
 }
+
+// FilterSliceElemsWithTestPredicate filters a slice using the test predicate function
+// and returns a slice that only contains elements that pass the test predicate function.
+func FilterSliceElemsWithTestPredicate(ss []string, testPredicate func(string) bool) (ret []string) {
```
Tatsinnit, line 367: 💡 `ss` - is the acronym for?

johnsonshi: I have renamed this to `slice`

deployment/aks-periscope.yaml
johnsonshi: Thanks for pointing this out! Removed and deleted.

Tatsinnit (approved):
👍 LGTM, thank you so much for this. 

I will defer to other maintainer folks for one last eyeballing, let’s merge it after that. Cc: @arnaud-tincelin since you’ve reviewed this before.  

Thanks 🙏

deployment/aks-periscope.yaml
arnaud-tincelin, line 31: This file is deprecated and shall not be updated anymore. These changes should be added to the corresponding kustomize resource file instead.
@Tatsinnit do you agree?

arnaud-tincelin (changes requested):

deployment/cluster-role.yaml
```diff
@@ -9,3 +9,9 @@ rules:
 - apiGroups: ["aks-periscope.azure.github.com"]
   resources: ["diagnostics"]
   verbs: ["get", "watch", "list", "create", "patch"]
+- apiGroups: ["apiextensions.k8s.io"]
+  resources: ["customresourcedefinitions"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["access.smi-spec.io", "specs.smi-spec.io", "split.smi-spec.io"]
+  resources: ["*"]
+  verbs: ["get", "list", "watch"]
```
johnsonshi, line 17: Hi @arnaud-tincelin, I've updated the kustomize resource file so that CRDs and SMI CRDs (both required) are added to the periscope cluster role permissions.
`deployment/cluster-role.yaml`

deployment/aks-periscope.yaml
johnsonshi: Hi @arnaud-tincelin, I think this should be resolved now that I've edited the corresponding kustomize resource file: https://github.com/Azure/aks-periscope/pull/71/files#r671532572

arnaud-tincelin (approved):

PR #75, opened 2021-06-28
WIP - support for KubernetesInDocker (kind) clusters

davidkydd:
Reverted previous premature merge (original PR here: https://github.com/Azure/aks-periscope/pull/69 )

test image available @ dakyddacr.azurecr.io/periscope/kind:v25


deployment/aks-periscope-kind.yaml
```diff
@@ -0,0 +1,191 @@
+# Note: this file is deprecated and will be removed in a future release
```
davidkydd, line 1: @Tatsinnit @arnaud-tincelin is the presence of this file the only outstanding concern with this PR? The other concerns raised on the previous version (https://github.com/Azure/aks-periscope/pull/69) were WRT logging format and usage of RunCommandOnHost - which are issues that I consider should be changed in separate PR's and consistently all-at-once rather than piecemeal. 

IMO this is fine to commit as is and we can refactor to a Kustomize version later. The reason I said "temp for development" on the previous PR was not intending that this be changed before merging, but that it would be eventually replaced with a Kustomize based version - just as the current aks-periscope.yaml will.

pkg/exporter/azureblob_exporter.go
```diff
@@ -19,23 +20,70 @@ const (
...
+	containerName = strings.TrimRight(containerName, "-")
+	return containerName, nil
+}
+
 // Export implements the interface method
 func (exporter *AzureBlobExporter) Export(files []string) error {
 	APIServerFQDN, err := utils.GetAPIServerFQDN()
 	if err != nil {
-		return err
+		return fmt.Errorf("Failed to get APIServerFQDN: %+v", err)
```
arnaud-tincelin, line 80: Do you have all tools for go? This line should have a warning
image
+ use `%w` in Errorf insted of %v
+ don't start with "fail to ...", as this message will be embedded in an Error message this adds no value

pkg/exporter/azureblob_exporter.go
```diff
@@ -19,23 +20,70 @@ const (
...
 	}
 
-	containerName := strings.Replace(APIServerFQDN, ".", "-", -1)
-	len := strings.Index(containerName, "-hcp-")
-	if len == -1 {
-		len = maxContainerNameLength
+	containerName, err := exporter.GetStorageContainerName(APIServerFQDN)
+	if err != nil {
+		return fmt.Errorf("Failed to get StorageContainerName: %+v", err)
```
arnaud-tincelin, line 85: same

pkg/exporter/azureblob_exporter.go
```diff
@@ -46,7 +94,7 @@ func (exporter *AzureBlobExporter) Export(files []string) error {
 	ses := utils.GetStorageEndpointSuffix()
 	url, err := url.Parse(fmt.Sprintf("https://%s.blob.%s/%s%s", accountName, ses, containerName, sasKey))
 	if err != nil {
-		return fmt.Errorf("Fail to build blob container url: %+v", err)
+		return fmt.Errorf("Failed to build blob container url: %+v", err)
```
arnaud-tincelin, line 97: `return fmt.Errorf("build blob container url: %w", err)`

pkg/exporter/azureblob_exporter_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+// TestGetAKSStorageContainerName get storage container name for non kind cluster
+func TestGetAKSStorageContainerName(t *testing.T) {
+	for _, tt := range getStorageContainerNameTests {
+		t.Run(tt.apiServerFqdn, func(t *testing.T) {
+			var blobExporter = &AzureBlobExporter{}
+			containerName, _ := blobExporter.GetAKSStorageContainerName(tt.apiServerFqdn)
+
+			if containerName != tt.containerName {
+				t.Errorf("Sprintf(%q, &blobExporter) => %q, want %q",
```
arnaud-tincelin, line 26: why use `Sprintf` when already using `Errorf`?

arnaud-tincelin (changes requested):
It's great we are adding some tests :)

pkg/utils/helper.go
```diff
@@ -52,6 +54,21 @@ func IsAzureStackCloud() bool {
 	return strings.EqualFold(cloud, AzureStackCloudName)
 }
 
+// IsRunningInAks returns true if the application is running on AKS
+func IsRunningInAks() bool {
+	//TODO refactor the conditional logic this check guards into new Cluster-Type specific components behind an interface
+	//test the non-AKS kubeconfig location, if we find something then this is *not* an AKS cluster
+	_, err := RunCommandOnHost("ls", "/etc/kubernetes/kubelet.conf")
```
Tatsinnit, line 61: 💡Why is this used to identify if the cluster is `AKS` or not?

cmd/aks-periscope/aks-periscope.go
```diff
@@ -32,40 +31,183 @@ func main() {
...
-	collectors = append(collectors, kubeObjectsCollector)
-	collectors = append(collectors, networkOutboundCollector)
 
-	if contains(collectorList, "connectedCluster") {
-		collectors = append(collectors, helmCollector)
+	//read list of collectors that are enabled
+	var enabledCollectorNames []string
+
+	//TODO try get partners to move from COLLECTOR_LIST to use ENABLED_COLLECTORS instead, for now COLLECTOR_LIST takes precedence if defined
+	collectorList, found := os.LookupEnv("COLLECTOR_LIST")
```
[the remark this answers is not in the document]
davidkydd: Nice! Proposal sounds good, second part might be tricky

pkg/collector/containerlogs_collector_containerd.go
```diff
@@ -0,0 +1,167 @@
...
+			filepath:      path.Join(directoryPath, logFile),
+		})
+	}
+
+	return containerLogs
+}
+
+//GetAllContainerLogFilesThatHaveEverRunOnHost gets the list of log files for all containers that have ever run on the host
+func (collector *ContainerLogsCollectorContainerD) GetAllContainerLogFilesThatHaveEverRunOnHost() ([]string, error) {
+	output, err := utils.RunCommandOnHost("ls", containerLogDirectory)
```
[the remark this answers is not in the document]
davidkydd: We should address removal of RunCommandOnHost all at once and as separate PR. Similar with windows which isn't currently supported anyway

pkg/exporter/azureblob_exporter_test.go
davidkydd: copy pasta - will fix

cmd/aks-periscope/aks-periscope.go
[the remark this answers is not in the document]
arnaud-tincelin: I see 2 basic options for now:
- do what the users was expected to do: check the cluster stack and decide whether to run or not
- always run, but return nothing if there is an error

pkg/collector/containerlogs_collector_containerd.go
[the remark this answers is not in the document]
arnaud-tincelin: The only reason it's not supported is this runcmd so let's at least limit the damage: we could leave legacy code as is but any new code shall not rely on these methods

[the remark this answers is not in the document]
davidkydd: Do you have ideas or suggestions for how to replace RunCommandOnHost? We can't call code that uses it legacy without a concrete plan for how to replace it, and I haven't myself come across anything suitable, or even confirmed that it an alternative is feasible.

Being able to run commands in the namespace of the host is one of the core functions of Periscope, and so "just this one function" is a big blocker - until this is clarified it seems unreasonable to halt any PRs on the basis that they leverage this method.

cmd/aks-periscope/aks-periscope.go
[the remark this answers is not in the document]
davidkydd: I disagree. For now we should retain the current behaviour - users might not want all collectors / diagnosers / exporters to execute, so even if we can confirm the current cluster stack and test that all c / d / e 's can run and safely err on every environment without side effect, it is useful to retain support for manual override.

COLLECTOR_LIST needs to be retained for backcompat with existing consumers. We can mark it deprecated though.

cmd/aks-periscope/aks-periscope.go
```diff
@@ -40,45 +40,192 @@ func main() {
...
+	runDiagnosers(diagnosers, exporters, diagnoserGrp)
+	diagnoserGrp.Wait()
+
+	zipAndExportString, found := os.LookupEnv("ZIP_AND_EXPORT")
+	zipAndExport, parseErr := strconv.ParseBool(zipAndExportString)
+	if !found || parseErr != nil {
+		zipAndExport = true
+	}
+
+	if zipAndExport {
```
arnaud-tincelin, line 68: The PR is about Kind cluster. Why all these changes in this file? What are we doing here?

cmd/aks-periscope/aks-periscope.go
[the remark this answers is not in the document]
arnaud-tincelin: it's just an env variable, we can drop it safely this won't break anything

cmd/aks-periscope/aks-periscope.go
```diff
@@ -40,45 +40,192 @@ func main() {
...
+
+	for _, diagnoserName := range enabledDiagnoserNames {
+		diagnosers = append(diagnosers, allDiagnosersByName[diagnoserName])
+	}
+
+	return diagnosers
+}
+
+// selectedExporters select the exporters to run
+func selectExporters(allExporters map[string]interfaces.Exporter) []interfaces.Exporter {
```
arnaud-tincelin, line 207: This is out of the PR scope

deployment/daemon-set.yaml
```diff
@@ -31,6 +31,12 @@ spec:
             name: kubeobjects-config
         - configMapRef:
             name: nodelogs-config
+        - configMapRef:
```
arnaud-tincelin, line 34: What's the link with Kind cluster?

pkg/collector/containerlogs_collector_containerd.go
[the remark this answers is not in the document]
arnaud-tincelin: Yeah the idea is to use adequate libraries : for kubectl use https://github.com/kubernetes/client-go, for helm use https://github.com/helm/helm/blob/main/pkg/kube/client.go, to access files we can mount volumes and use io / ioutils libraries, etc ...
I'm not saying to remove RunCmdOnHost now but aks-periscope is not aimed to be a wraper to call cli tools. We have libraries to do it properly so let's do it :)
Leave old code as is but for new code, let's use the libraries, especially regarding files

pkg/exporter/azureblob_exporter.go
```diff
@@ -32,48 +33,8 @@ func NewAzureBlobExporter(creationTime, hostname string) *AzureBlobExporter {
 	}
 }
 
-func createContainerURL() (azblob.ContainerURL, error) {
```
arnaud-tincelin, line 35: why the exporter changes?

arnaud-tincelin (changes requested):
Sorry but I'm lost in this PR :(

pkg/collector/containerlogs_collector_containerd.go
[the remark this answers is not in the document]
davidkydd: Ahhh right ok will try with volume mount, thanks

cmd/aks-periscope/aks-periscope.go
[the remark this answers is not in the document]
davidkydd: It won't break anything for us but will force changes on partners, which I have been trying to minimize. It is also self contained and easy to remove later

pkg/utils/helper.go
Tatsinnit: Yeah,  This could be done more like fall back checks which will help in better, failure error once all fails for `non-aks` scenario.

I would highly recommend this:

1. check for cluster-info : using `kubectl cluster-info` 
    *  For AKS it will be `-aks-fqdn.azmk8s.io.` , and other providers have similar patterns.
2. Also, If we run with in the cluster then we could have used :https://kubernetes.io/docs/tasks/administer-cluster/namespaces/#deleting-a-namespace 
3. check third fallback and then -> graciously fail with right messages.

It is key that we right these functions correctly so the we can identify what all kind cluster we could be extending this support.

cc: @safeermohammed , @arnaud-tincelin or @palma21 - for more ideas to piggy back on this.

cmd/aks-periscope/aks-periscope.go
```diff
@@ -120,36 +260,31 @@ func main() {
...
+	var result error
+	for _, e := range exporters {
+		if err := e.Export(dataProducer); err != nil {
+			result = multierror.Append(result, err)
 		}
 	}
-
-	// TODO: Hack: for now AKS-Periscope is running as a deamonset so it shall not stop (or the pod will be restarted)
-	// Revert from https://github.com/Azure/aks-periscope/blob/b98d66a238e942158ef2628a9315b58937ff9c8f/cmd/aks-periscope/aks-periscope.go#L70
-	select {}
```
Tatsinnit, line 142: 💡side note:  See the comment above, this is needed at its current state. Unless something is changed which fixes this.

cmd/aks-periscope/aks-periscope.go
```diff
@@ -120,36 +260,31 @@ func main() {
...
-
-	// TODO: remove this //nolint comment once the select{} has been removed
-	//nolint:govet
+	return result
 }
 
-func contains(flagsList []string, flag string) bool {
```
Tatsinnit, line 148: 💡Key: Not sure what are these changes, and what was the intention of removing this function?

pkg/exporter/azureblob_exporter.go
Tatsinnit, line 35: 💡 How these changes relates with the Kind cluster support which this PR is intending? Is this just moved below or what exactly is changed.

pkg/utils/helper.go
```diff
@@ -103,28 +120,59 @@ func GetHostName() (string, error) {
...
-	}
-
+//ParseAPIServerFQDNFromKubeConfig parses a kubeConfig and returns the APIServerFQDN
+func ParseAPIServerFQDNFromKubeConfig(output string) (string, error) {
 	lines := strings.Split(output, "\n")
 	for _, line := range lines {
 		index := strings.Index(line, "server: ")
 		if index >= 0 {
 			fqdn := line[index+len("server: "):]
-			fqdn = strings.Replace(fqdn, "https://", "", -1)
```
Tatsinnit, line 119: 🐛 test for private cluster, this was specific for that. (I think)

pkg/utils/helper.go
```diff
@@ -312,6 +360,16 @@ func GetResourceList(kubeCmds []string, separator string) ([]string, error) {
 	return strings.Split(strings.Trim(resourceList, "\""), separator), nil
 }
 
+//contains checks if an array contains a value
+func Contains(flagsList []string, flag string) bool {
```
Tatsinnit, line 364: 💡 Not able to follow this moving around of code because previously this funct was privately used in the main `aks-perisceop.go` file

Tatsinnit (changes requested):
💡 Thanks for this,I will add bit more latter. This PR is more then adding Kind support as it looks and some of the code movement around is something I am still going through. 

* Regarding supporting the actual kind cluster, we should implemented well thought mechanism detailed here: https://github.com/Azure/aks-periscope/pull/75/files#r662950855

cc: @arnaud-tincelin, @JunSun17 - for more thoughts on other changes. 

🙏 Thank you

davidkydd (commented):
> Sorry but I'm lost in this PR :(

It might be easier to review just the raw code itself outside of diff form by checking out the branch the PR is on.

I usually find this easier with larger PR's which change lots of bits of files due to refactoring, e.g. I had to do this recently for your PR https://github.com/Azure/aks-periscope/pull/65

Tatsinnit:
💡 Is this PR ready for review? 

Important note: Kubernetes in docker is deprecated now, is the key purpose of this PR will be affected by this?

* FYI: https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker/ 

Thanks,

davidkydd:
> 💡 Important note: Kubernetes in docker is deprecated now, is the key purpose of this PR will be affected by this?
> 
> * FYI: https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker/
> 
> Thanks,

Hey, no this PR isn't affected - also its not quite right to say that "Kubernetes in Docker is being deprecated", as its the dockershim container runtime which is being deprecated and replaced with containerd. KIND certainly isn't being deprecated (see banner at the tope of https://kind.sigs.k8s.io/). 

As this PR adds a containerd based containerLog collector you could say this PR is even MORE applicable considering this deprecation 😄

pkg/collector/containerlogs_collector_containerd.go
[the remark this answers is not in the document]
davidkydd: FYI the change to use a volume mount has forced a refactor to target the logs in /var/log/pods/ not /var/log/containers/, as /containers/ just contains symlinks to the actual log files in /pods/ :doh:

Tatsinnit (changes requested):
💡 Thank you for this, here are some observations:

* This PR is not just Specific k8sindocker support **but**  3 other things, i.e. `containerd`, `some local machine support` along with `code move around` hence it's little confusing to follow the flow of logic. 
    * If we divide this into smaller feature based PR we might be able to understand better.
        * Especially, LocalMachine mechanism, and specifics of k8sindocker support.
* Might be split into smaller PR: I agree with initial comments of @arnaud-tincelin that it seems like this PR is not just `Support for some specific Kind` but became more then 3 features in 1. It will be much easy to spilt into smaller more chunks for better understanding. Unless it cannot be separated.
* For tests I would recommend `Gomega` which is already used within this project and much better assertion notation.

I will defer to @arnaud-tincelin for further comments, suggestions or recommendation.

Thank you. 🙏

pkg/exporter/azureblob_exporter.go
arnaud-tincelin: ok, cool. I'd suggest to move this whole thing to another PR

cmd/aks-periscope/aks-periscope.go
arnaud-tincelin: I understand but right now the PR is hard to review and you have to deal with our numerous comments.
I think exporters shall not be treated the same way as collectors and diagnosers: exporters shall be configured by the user (export to blob, to disk, ...) but as we already discussed, I believe collectors and diagnosers should be 'hidden' from the users.

May be before a PR to support Kind cluster we should address these points in a dedicated PR ?

cmd/aks-periscope/aks-periscope.go
arnaud-tincelin: Let's have 3 PR:
- configuration of exporters + mechanism to run (or not) collectors / diagnosers
- PR for zip (not sure what we need to do here)
- PR to support kind cluster: if others PR are done, this one shall be very small

cmd/aks-periscope/aks-periscope.go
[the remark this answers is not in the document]
arnaud-tincelin: I don't see the point with issue here. The worst it can do is gather more files, the backcompat you are talking about is ensured. There is no change needed for the partners.
Let's tackle this in another PR if you want

cmd/aks-periscope/aks-periscope.go
arnaud-tincelin: ZipAndExport step shall not exist. We will remove it asap but I know other tools such as vscode are using it.
This zip step is a duplicate of the single exports per collector / diagnoser

pkg/collector/containerlogs_collector_containerd.go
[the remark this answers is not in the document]
arnaud-tincelin: This is the reason why we mount /var/log:
```yaml
 volumes:
      - name: varlog
        hostPath:
          path: /var/log
```

arnaud-tincelin (changes requested):
I think we should split this PR as it's taking too long to review, Tatsat and I are getting lost about the reasons behind several changes and we disagree on some points

pkg/utils/helper.go
```diff
@@ -103,28 +106,57 @@ func GetHostName() (string, error) {
...
 	return "", errors.New("Could not find server definitions in kubeconfig")
 }
 
+//ReadKubeletConfig reads the kubeletConfig from the node
+func ReadKubeletConfig() (string, error) {
+	var result error
+	output, err := RunCommandOnHost("cat", "/var/lib/kubelet/kubeconfig")
+	if err != nil {
+		result = multierror.Append(result, err)
+		output, err = RunCommandOnHost("cat", "/etc/kubernetes/kubelet.conf")
```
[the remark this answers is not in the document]
sophsoph321: @Tatsinnit, the reason for using /etc/kubernetes/kubelet.conf is to override the hardcoded value /var/lib/kubelet/kubeconfig. For kind and kubeadm distros, we will get a "can't read kubeconfig file" if we try to get the APIServerFQDN from the filepath /var/lib/kubelet/kubeconfig.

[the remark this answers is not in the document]
Tatsinnit: Yep, @sophsoph321, thanks, Indeed, that is why as recommended, I think it is important to leave a comment i.e. add a code level comment for this: `commenting the reason` this config is use, **especially** when specific code if for k8sindocker (KubeKDM) kind it is not every kind Kubernetes cluster.

For anyone to look at the code it will make it clear that what all this change was specifically for.

nicoletaz3:
Hi Tatsat, Docker is deprecated as a container runtime in Kubernetes, but not KIND which is a tool for running Kubernetes in a lightweight manner. 

---
In reply to: [874466230](https://github.com/Azure/aks-periscope/pull/75#issuecomment-874466230) [](http://example.com/codeflow?ancestors=874466230,874466230)

Tatsinnit:
> Hi Tatsat, Docker is deprecated as a container runtime in Kubernetes, but not KIND which is a tool for running Kubernetes in a lightweight manner.
> 
> In reply to: [874466230](https://github.com/Azure/aks-periscope/pull/75#issuecomment-874466230)

Thank you @nicoletaz3 , yep sounds good. Really appreciate your reply, As a archival history I left this question since there is no other doc but this PR for this work. 🙏

Regarding this PR: There are bunch of key comments which are getting checked in as part of this PR and some are not clear from hence Arnaud’s and mine comments above in case you are reviewing this PR please feel free to add comments. Thank you so much.

davidkydd:
Closing - will be refactored into multiple smaller PRs

pkg/collector/containerlogs_collector_containerd.go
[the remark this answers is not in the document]
davidkydd: yes, however we need to change the code significantly to read from /var/log/pods rather than /var/log/containers as the directory structure is different. With /var/log mounted it still wouldn't follow the symlink from /containers to /pods

PR #80, opened 2021-07-08
update readme

qpetraroia:
Quick PR to update the readme file for aks-periscope. We want to tell the users that `az-cli` and` vscode` have a dependency on the current `aks-persicope.yaml` file. It is important that they know that. As breaking changes can occur if they edit it.


Tatsinnit (approved):
👍 Thank you for this @qpetraroia , Looks good to me, and probably more clear for the contract of usage between this tool and other 2 tools. I will approve and will wait until other folks have anything else to add. 

Thank you. 🙏

Issue #81, opened 2021-07-11
Diagnoser WriteToCrd broken during recent refactor


davidkydd:
**Describe the bug**
Diagnoser WriteToCrd broken in https://github.com/Azure/aks-periscope/commit/cce800065d8835cb1588a48a0176967e45d54cb2 

**How bug surfaces**
When running periscope, diagnosers fail to update the CRD, logging shows:
2021/07/09 04:19:54 Diagnoser: networkoutbound, diagnose data failed: write data from NetworkOutbound Diagnoser to CRD: open [{"HostName":"aks-agentpool-14754340-vmss000000","Type":"AKS API Server","Start":"2021-07-09T04:17:24Z","End":"2021-07-09T04:17:24Z","Status":"Connected"},{"HostName":"aks-agentpool-14754340-vmss000000","Type":"AKS Tunnel","Start":"2021-07-09T04:17:24Z","End":"2021-07-09T04:17:24Z","Status":"Error: dial tcp 20.62.238.205:9000: connect: connection refused"},{"HostName":"aks-agentpool-14754340-vmss000000","Type":"Azure Container Registry","Start":"2021-07-09T04:17:24Z","End":"2021-07-09T04:17:24Z","Status":"Connected"},{"HostName":"aks-agentpool-14754340-vmss000000","Type":"Microsoft Container Registry","Start":"2021-07-09T04:17:24Z","End":"2021-07-09T04:17:24Z","Status":"Connected"},{"HostName":"aks-agentpool-14754340-vmss000000","Type":"Internet","Start":"2021-07-09T04:17:24Z","End":"2021-07-09T04:17:24Z","Status":"Connected"}]: file name too long
2021/07/09 04:19:54 Diagnoser: networkconfig, diagnose data failed: write data from NetworkConfig Diagnoser to CRD: open {"HostName":"aks-agentpool-14754340-vmss000000","NetworkPlugin":"","VirtualMachineDNS":null,"KubernetesDNS":["10.0.0.10\nsearch"],"MaxPodsPerNode":0}: no such file or directory

**Additional context**
Fix for Arc fork can be seen here: https://github.com/davidkydd/aks-periscope/commit/ba215e55e606bfa7df5411da74da34641547b403

PR #82, opened 2021-07-13
fix for CRD. Only read hostname once and pass it around instead

davidkydd:
this fixes the bug introduced here: https://github.com/Azure/aks-periscope/commit/cce800065d8835cb1588a48a0176967e45d54cb2

and the issue reported here:
https://github.com/Azure/aks-periscope/issues/81

Bug was introduced during the move to not use files, but the WriteToCRD function still attempted to read from the diagnoser output file

### DIFF
--- a/cmd/aks-periscope/aks-periscope.go
+++ b/cmd/aks-periscope/aks-periscope.go
@@ -100,8 +100,8 @@ func main() {
 	collectorGrp.Wait()
 
 	diagnosers := []interfaces.Diagnoser{
-		diagnoser.NewNetworkConfigDiagnoser(dnsCollector, kubeletCmdCollector),
-		diagnoser.NewNetworkOutboundDiagnoser(networkOutboundCollector),
+		diagnoser.NewNetworkConfigDiagnoser(hostname, dnsCollector, kubeletCmdCollector),
+		diagnoser.NewNetworkOutboundDiagnoser(hostname, networkOutboundCollector),
 	}
 
 	diagnoserGrp := new(sync.WaitGroup)
--- a/pkg/utils/helper.go
+++ b/pkg/utils/helper.go
@@ -251,22 +251,13 @@ func GetCreationTimeStamp() (string, error) {
 }
 
 // WriteToCRD writes diagnostic data to CRD
-func WriteToCRD(fileName string, key string) error {
-	hostName, err := GetHostName()
-	if err != nil {
-		return err
-	}
+func WriteToCRD(jsonBytes []byte, key string, hostName string) error {
 
 	crdName := "aks-periscope-diagnostic" + "-" + hostName
 
-	jsonBytes, err := ioutil.ReadFile(fileName)
-	if err != nil {
-		return err
-	}
-
 	patchContent := fmt.Sprintf("{\"spec\":{%q:%q}}", key, string(jsonBytes))
 
-	_, err = RunCommandOnContainer("kubectl", "-n", "aks-periscope", "patch", "apd", crdName, "-p", patchContent, "--type=merge")
+	_, err := RunCommandOnContainer("kubectl", "-n", "aks-periscope", "patch", "apd", crdName, "-p", patchContent, "--type=merge")
 	if err != nil {
 		return err
 	}
--- a/pkg/diagnoser/networkconfig_diagnoser.go
+++ b/pkg/diagnoser/networkconfig_diagnoser.go
@@ -20,14 +20,16 @@ type networkConfigDiagnosticDatum struct {
 
 // NetworkConfigDiagnoser defines a NetworkConfig Diagnoser struct
 type NetworkConfigDiagnoser struct {
+	hostName            string
 	dnsCollector        *collector.DNSCollector
 	kubeletCmdCollector *collector.KubeletCmdCollector
 	data                map[string]string
 }
 
 // NewNetworkConfigDiagnoser is a constructor
-func NewNetworkConfigDiagnoser(dnsCollector *collector.DNSCollector, kubeletCmdCollector *collector.KubeletCmdCollector) *NetworkConfigDiagnoser {
+func NewNetworkConfigDiagnoser(hostName string, dnsCollector *collector.DNSCollector, kubeletCmdCollector *collector.KubeletCmdCollector) *NetworkConfigDiagnoser {
 	return &NetworkConfigDiagnoser{
+		hostName:            hostName,
 		dnsCollector:        dnsCollector,
 		kubeletCmdCollector: kubeletCmdCollector,
 		data:                make(map[string]string),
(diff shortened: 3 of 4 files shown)


cmd/aks-periscope/aks-periscope.go
```diff
@@ -100,8 +100,8 @@ func main() {
 	collectorGrp.Wait()
 
 	diagnosers := []interfaces.Diagnoser{
-		diagnoser.NewNetworkConfigDiagnoser(dnsCollector, kubeletCmdCollector),
-		diagnoser.NewNetworkOutboundDiagnoser(networkOutboundCollector),
+		diagnoser.NewNetworkConfigDiagnoser(hostname, dnsCollector, kubeletCmdCollector),
```
davidkydd, line 103: strictly speaking this change isn't needed to fix the bug - but by passing the hostName we can remove two additional calls to GetHostName(), which currently requires running external cmd lines

Tatsinnit: 💡 Reducing the parameters is a good thing, I believe adding new parameter will make dependable function hence removes the “single responsibility”

Having one function like getHostName() helps in calling whenever it is needed. Unless I am missing a key advantage I believe we should have lesser deep parameter passing.

davidkydd: But then everything ends up tightly coupled to a giant "helpers" file full of utils methods like we have currently? This approach scales poorly and (most importantly) limits what we can do from unit tests.

Adding this parameter is not removing the single responsibility or adding a secondary function, its just making it explicit that the diagnoser and utils.WriteToCRD require access to a piece of state (the hostname) in order to perform their single responsibility. 

If we really wanted to over-engineer this we could abstract both utils methods in question (GetHostName and WriteToCRD) into their own types behind interfaces, give the WriteToCRD an implementation of GetHostName, and then give each diagnoser a WriteToCRD implementation.

Tatsinnit: 💡 Having single responsible functions which are easily testable or much more maintainable is better then dependable and more weaved functions.

Adding more parameter is not good especially the use in this specific this case when it’s easily doable. Essentially this could be done more simply like this: 
* WriteToCRD —> Already had `hostName, err := GetHostName()` in sue why we are deliberately removing this simplicity and passing this all the way down through Parameters seems convolution to me? (Unless I am missing a key point)

Having bigger helper file is not an issue because we can always divide it into separate classes according to their responsibilities and add test. and We can always have Utils/helper to more accurate classes but having good helper method which are easily testable are much better and much easy separation of concern.

cc: @arnaud-tincelin as well. 

* Honestly the fix, what I can see from the changes in this PR is just remove line 262 through 265 in helper.go?

Tatsinnit (changes requested):
💡  Having single responsible functions which are easily testable or much more maintainable is better then dependable and more weaved functions. 

What part of this PR is doing is Passing parameter which was **already** simply used in the lower level function, see here https://github.com/Azure/aks-periscope/blob/master/pkg/utils/helper.go#L255 so instead of this, this PR will add noise via extra parameter which is not used anywhere but at this lower level funct level?

Adding more parameter is not good especially the use in this specific case when it’s easily doable. Essentially this could be done more simply like this:

* WriteToCRD —> Already had hostName, err := GetHostName() in sue why we are deliberately removing this simplicity and passing this all the way down through Parameters seems convolution to me? (Unless I am missing a key point)

* Honestly the fix, what I can see from the changes in this PR is just remove line 262 through 265 in helper.go?

cc: @arnaud-tincelin as well for review.

cmd/aks-periscope/aks-periscope.go
davidkydd: _ Having single responsible functions which are easily testable or much more maintainable is better then dependable and more weaved functions._ both of the "pure" utils methods are unchanged, they are still single responsible functions which are easily testable and just as maintainable as before. 

_Adding more parameter is not good especially the use in this specific this case when it’s easily doable_ you are over generalizing - sometimes adding more parameters is the best approach, such as I argue is the case here.

_WriteToCRD —> Already had hostName, err := GetHostName() in sue why we are deliberately removing this simplicity and passing this all the way down through Parameters seems convolution to me? (Unless I am missing a key point)_  The current usage of GetHostName is bad, it leads to an expensive call:

hostname, err := RunCommandOnHost("cat", "/etc/hostname")

which is made at various times all over the program unnecessarily as the hostName will never change during program execution (the method is pure), instead, this expensive call should only be made once and the value somehow made available to the parts of the program which need to access it.

The other concern is making sure that this value can be modified and injected during unit testing of those parts of the program which use it - one way of doing that is by passing it as a parameter (the one I have chosen here) the other way is by abstracting it behind an interface (the approach I mentioned in my first reply, considered over-engineering at this stage)

_Having bigger helper file is not an issue because we can always divide it into separate classes according to their responsibilities and add test. and We can always have Utils/helper to more accurate classes but having good helper method which are easily testable are much better and much easy separation of concern._ tight coupling to helper methods *not* behind an interface means that we cannot change the behaviour of the helper methods when we want to test the things that are calling them (like the diagnoser in this case). 

What you are describing is not separation of concerns; because of the tight coupling, the concerns of the helper functions are being directly added to the concerns of the component which calls them (the diagnoser).

Tatsinnit: 💡Cool, I think above is copy paste paste and para reply so hard to follow: so might be its good to check that Isn't this the only fix needed for this bug: https://github.com/Tatsinnit/aks-periscope/commit/1f9ab3f9a19d9b410bf5a4da3008983db0f9ea26 please correct me?

davidkydd: Sure let's split it then, do you want to open a PR of your commit above (just the fix) and we can address all the extraneous calls to GetHostName in a separate refactor PR

Tatsinnit: + 1 Thank you, sound good.

arnaud-tincelin: I didn't read all of this discussion but I think for now it's the best option

pkg/utils/helper.go
```diff
@@ -251,22 +251,13 @@ func GetCreationTimeStamp() (string, error) {
 }
 
 // WriteToCRD writes diagnostic data to CRD
-func WriteToCRD(fileName string, key string) error {
-	hostName, err := GetHostName()
-	if err != nil {
-		return err
-	}
+func WriteToCRD(jsonBytes []byte, key string, hostName string) error {
```
arnaud-tincelin, line 254: not the point of this PR but why do we need this function?

arnaud-tincelin (approved):

pkg/utils/helper.go
davidkydd: "diagnostics" values are exposed via a k8s custom resource created under the periscope api namespace (see helper.go  createCRD() and writeDiagnosticCRD())

cmd/aks-periscope/aks-periscope.go
davidkydd: @Tatsinnit I see Arnaud has approved - what do you want to do here? Should I merge this as is or split into a fix and a refactor PR

Tatsinnit: I think he is agreeing with what I said , @arnaud-tincelin for clarity.

What he mean is that we should have separate PR for refactor and a bug fix, also I think he was wondering why there are no read operation but that is separate thing. Thanks.

PR #83, opened 2021-07-13
Selectable C, D, E's

davidkydd:
Replace COLLECTOR_LIST config with ENABLED_COLLECTORS, ENABLED_DIAGNOSERS, ENABLED_EXPORTERS.

Basically this works by creating a type-registry, by instantiating a map of type-names -> instance of the type, and then filters the map against a []string, so that only types with names which are present in the []string are selected for execution.

There are some alternative (and more comprehensive) ideas for a type registry here: https://stackoverflow.com/questions/23030884/is-there-a-way-to-create-an-instance-of-a-struct-from-a-string

I think "dolmen"s answer looks good, as it would let us get rid of the "GetName()" method and just use the name of the struct directly, whilst also allowing us to only instantiate the types we end up executing. But I feel that could probably be done as a later refactor.


cmd/aks-periscope/aks-periscope.go
```diff
@@ -54,31 +104,103 @@ func main() {
...
-		collectors = append(collectors, osmCollector)
+	return collectors
+}
+
+// selectDiagnosers select the diagnosers to run
+func selectDiagnosers(allDiagnosersByName map[string]interfaces.Diagnoser) []interfaces.Diagnoser {
+	diagnosers := []interfaces.Diagnoser{}
+
+	//read list of diagnosers that are enabled
+	enabledDiagnoserString, found := os.LookupEnv("ENABLED_DIAGNOSERS")
```
arnaud-tincelin, line 166: Are ENABLED_XXX new variables? If yes, is it possible to remove them completely?
We already discussed this point and I think for a first iteration, having
		//if not defined, default to all diagnosers enabled
		enabledDiagnoserString = "networkconfig networkoutbound"
is a what we want.
The user should only have to choose exporters

What do you think? @Tatsinnit

pkg/exporter/azureblob_exporter.go
```diff
@@ -32,6 +32,10 @@ func NewAzureBlobExporter(creationTime, hostname string) *AzureBlobExporter {
 	}
 }
 
+func (exporter *AzureBlobExporter) GetName() string {
```
arnaud-tincelin, line 35: Why do we need GetName for exporters?

arnaud-tincelin (changes requested):

cmd/aks-periscope/aks-periscope.go
davidkydd: Yes new variables - yes we discussed this and I argue that it is a useful capability to support across all collectors / diagnosers / exporters. Arc need collectors to be configurable like this also, and it is only a matter of time until diagnosers need to be as well. No harm in making them default to all on (the user doesn't **have** to choose) but also support overriding like this.

pkg/exporter/azureblob_exporter.go
davidkydd: so we can use it to select them, currently only one exporter (azureblob) but I have PR in the wings for localmachine as well.

cmd/aks-periscope/aks-periscope.go
Tatsinnit: I will plus 1 to **not** adding new variables without thinking whole causal effect it will have, especially some of the `osm`, `smi` work is utilising `collector_list` to uniquely identify new work.

* Currently the `collectors/diagnosers` are set of default periscope collectors, as the OSM and SMI (Service Mesh team) added we have added this logic https://github.com/Azure/aks-periscope/blob/master/cmd/aks-periscope/aks-periscope.go#L76 to handle those scenarios.

How adding new variable is needing clarity in this design:

* Breaking change as to what is the default working behaviour?
* Why are we adding these new variable? What is the use case, what requirement needs it as `selectable` option? 

Thanks 🙏

cmd/aks-periscope/aks-periscope.go
```diff
@@ -55,35 +105,104 @@ func main() {
...
+		})
+
+	return selectedCollectors, selectedDiagnosers, selectedExporters
+}
+
+// selectCollectors select the collectors to run by looking up the value from the config
+func selectCollectors(allCollectorsByName map[string]interfaces.Collector) []interfaces.Collector {
+	collectors := []interfaces.Collector{}
+
+	enabledCollectorString, found := os.LookupEnv("ENABLED_COLLECTORS")
```
Tatsinnit, line 148: 💡 What will be associated `deployment.yaml` changes will look like? For this new var `ZIP_AND_EXPORT`

cmd/aks-periscope/aks-periscope.go
```diff
@@ -55,35 +105,104 @@ func main() {
...
+	if !found {
+		//if not defined, default to all collectors enabled
+		enabledCollectorString = "containerlogs dns helm iptables kubeletcmd kubeobjects networkoutbound nodelogs osm smi systemlogs systemperf"
```
Tatsinnit, line 151: 💡 If this is `default list` of collectors, then this is incorrect because `osm, smi` both are by conditional check checks, which are based one `OSM` as value in collectorList

cmd/aks-periscope/aks-periscope.go
```diff
@@ -55,35 +105,104 @@ func main() {
...
+
+	return diagnosers
+}
+
+// selectedExporters select the exporters to run
+func selectExporters(allExporters map[string]interfaces.Exporter) []interfaces.Exporter {
+	exporters := []interfaces.Exporter{}
+
+	//read list of exporters that are enabled
+	enabledExportersString, found := os.LookupEnv("ENABLED_EXPORTERS")
```
Tatsinnit, line 188: 💡What will be associated `deployment.yaml` changes will look like? For this new var.

cmd/aks-periscope/aks-periscope.go
```diff
@@ -40,8 +42,56 @@ func main() {
...
+
+	collectorGrp := new(sync.WaitGroup)
+	runCollectors(collectors, exporters, collectorGrp)
+	collectorGrp.Wait()
+
+	diagnoserGrp := new(sync.WaitGroup)
+	runDiagnosers(diagnosers, exporters, diagnoserGrp)
+	diagnoserGrp.Wait()
+
+	zipAndExportString, found := os.LookupEnv("ZIP_AND_EXPORT")
```
Tatsinnit, line 64: 💡What will be associated `deployment.yaml` changes will look like? For this new var `ZIP_AND_EXPORT`

cmd/aks-periscope/aks-periscope.go
davidkydd: CollectorList value of OSM currently selects both OSM and SMI collectors to run. Value of SMI selects just SMI collector. Intention is for default behaviour being to run ALL collectors

cmd/aks-periscope/aks-periscope.go
davidkydd: Correct, COLLECTOR_LIST is intended to be deprecated. There are no backwards compatibility issues or breaking changes with removing it as far as I am aware, as long as azCLI and VSCode extension haven't yet changed to use it? 

Arc scenario requires being able to only run some collectors, not the full set. Values will be provided via deployment yaml, example here: https://github.com/davidkydd/aks-periscope/blob/0071201dd41cd159e87e5a574bdfbcc271c122b1/deployment/aks-periscope-arc.yaml#L147

cmd/aks-periscope/aks-periscope.go
davidkydd: https://github.com/davidkydd/aks-periscope/blob/0071201dd41cd159e87e5a574bdfbcc271c122b1/deployment/aks-periscope-arc.yaml#L163

cmd/aks-periscope/aks-periscope.go
davidkydd: https://github.com/davidkydd/aks-periscope/blob/0071201dd41cd159e87e5a574bdfbcc271c122b1/deployment/aks-periscope-arc.yaml#L164 - minor nitpick but if you want to try find a way that the bool value doesn't have to be wrapped in double quotes that would be preferable

cmd/aks-periscope/aks-periscope.go
davidkydd: for current value of collectorlist = OSM see https://github.com/Azure/aks-periscope/blob/a1bdc52496b5c05547236e2cc7c4d3b15cd1359f/cmd/aks-periscope/aks-periscope.go#L76

davidkydd: Well, see the two lines under that one 🙂

cmd/aks-periscope/aks-periscope.go
davidkydd: https://github.com/davidkydd/aks-periscope/blob/0071201dd41cd159e87e5a574bdfbcc271c122b1/deployment/aks-periscope-arc.yaml#L147

cmd/aks-periscope/aks-periscope.go
davidkydd: We should try remove the need to update this string as a requirement of adding a new collector. Should dovetail well with the change to use a proper type registry mentioned in initial PR description.

cmd/aks-periscope/aks-periscope.go
Tatsinnit: 💡 `Adding this as a thought` : So, adding new configurable `vars`, will become a breaking change for **az-cli and vscode** as well because those `default vars` need to be maintained somehow in most cleanest non scattered manner (not sure how). (some proposal was here) I will think more about this **but** _**I think at this point see this discussion here**_: https://github.com/Azure/aks-periscope/pull/75/files#r661261039 

* Essentially, in current state when we remove the "Collector_List", **and** adding new `vars` it will break things, and any new changes on top will be backward breaking change,  Thanks.

davidkydd: My fork retains support for COLLECTOR_LIST for backcompat while adding support for SELECTED_COLLECTORS too. COLLECTOR_LIST is checked first and used preferentially if both are defined. If you want I can bring this code to the PR too, then we can deprecate COLLECTOR_LIST and move VSCode / CLI to use SELECTED_COLLECTORS at our leisure before removing COLLECTOR_LIST support from core periscope?

arnaud-tincelin: I don't see this issue with backcompat since if we removed this the log would still be collected as before (with some more, but this can't be an issue)

pkg/exporter/azureblob_exporter.go
arnaud-tincelin: I do hope we are not going to select any structure based on a display name :D

cmd/aks-periscope/aks-periscope.go
arnaud-tincelin: I'd suggest not to modify this part as we should remove this zip function asap. What do you think?

cmd/aks-periscope/aks-periscope.go
Tatsinnit: 💡Also, Going back and possibly asking again why are we adding this again, please? `ENABLED_COLLECTORS, ENABLED_DIAGNOSERS, ENABLED_EXPORTERS.` 

* Is this `ARC` specific requirement which is the key scenario here or is this the refactor?

also cc: @arnaud-tincelin in case you know. Thanks.

cmd/aks-periscope/aks-periscope.go
Tatsinnit: 💡 +1 to what Arnaud is trying to convey, I think `ZIP_AND_EXPORT` is a new design somehow (or Use case probably), like mentioned above when `true` what does it achieve? 

* i.e. in current already re-factored `in-memory` files get uploaded as logs, and `zipped` in the end.

This new flag adds new behaviour. Why is this flag required at all `ZIP_AND_EXPORT`?

cmd/aks-periscope/aks-periscope.go
davidkydd: @arnaud-tincelin the backcompat issue is to do with integration with az-cli and vscode extensions, and would prevent the issue which @Tatsinnit described here being a breaking change: https://github.com/Azure/aks-periscope/pull/83#discussion_r675228034

Yes ENABLED_COLLECTORS is a direct replacement for COLLECTOR_LIST - Arc also have a requirement for a LocalMachineExporter and we need a way to selectively enable and disable this functionality, 

The ability to determine which components (collectors, diagnosers, exporters) are executed via the deployment yaml is a generally useful feature to support, as it allows the user to control the behaviour of periscope at a coarse but clearly defined level. We have already seen from just a few feature requests that it is necessary to be able to control which collectors and exporters are executed, so we should be proactive and enable this capability for diagnosers at the same time. 

Therefore:
We should standardize on naming and either add the three configs:

ENABLED_COLLECTORS
ENABLED_EXPORTERS 
ENABLED_DIAGNOSERS 

or the equivalent:

COLLECTOR_LIST
EXPORTER_LIST
DIAGNOSER_LIST

The form ENABLED_XYZ  is more explicit than XYZ_LIST as to what it actually does, which is why I am suggesting we use the ENABLED_XYZ form - imagine later on if we want to invert this logic for some reason (a very plausible scenario) and add a DISABLED_COLLECTORS config flag, for the requirement that an environment can define what collectors *cannot* run on it, but so that it will automatically include and run any new collectors that are added by default. 

In this scenario COLLECTOR_LIST would become ambiguous as to its function (is it the list of collectors that are enabled, or disabled?)

pkg/exporter/azureblob_exporter.go
davidkydd: why not? This is perfect for the interim - as I mentioned in initial description, a better idea is to use a type registry and remove the need for GetName to be defined at all, but this is overkill for now.

Using the display name means we are not burdened with the unnecessary complexity of introducing an intermediate concept whose only purpose is to map from one name to another, and which we would have to refactor to remove if and when we move to a type registry anyway.

cmd/aks-periscope/aks-periscope.go
davidkydd: 1. The benefit of removing the zip function is that there is less processing done on the node running periscope.
2. The benefit of **using*** the zip function is that there are less bits needed to be sent over the network.

I agree that the current usage of the zip functionality is not optimized. However as there are benefits in both zipping and not-zipping, instead of removing it we should make it configurable.

Zipping has less benefit when exporting to AzureBlob, because the traffic will usually all be within the MSFT network and hence size is less critical (though still important). 

Arc have asked for a LocalMachineExporter, and the CLI is then used to download periscope outputs directly from the node. This should be zipped first as size is likely more important consideration.

The current function of ZIP_AND_EXPORT is a minimal change which supports the Arc LocalMachineExporter scenario, and allows other clients to still disable the ZIP functionality as an optimization if they wish to just rely on the current behaviour of the AzureBlobExporter. 

Later - we should refactor the AzureBlobExporter so that it also respects the ZIP_AND_EXPORT config (or some equivalent config), which will allow the user to configure the behaviour, and decide which of 1. or 2. is more important for their scenario. In memory files should either be zipped and exported to AzureBlob, or just exported as files - doing both is redundant. The ZIP_AND_EXPORT setting (or whatever we decide to call it) should be configurable for each Exporter, so that users can if they wish, have the AzureBlobExporter exporting un-zipped files, and the LocalMachineExporter exporting a zipped file. Zipping should only be performed once and the file shared by exporters for efficiency.

Tatsinnit: 💡 So this refactor is based on some kind of requirements? Thats where we think there is something neither Arnaud and I know or understand, hence the point fo this whole change is not convoluted.

* Currently this tool ==> Runs as daemonset -> collects logs (collector/diagnoser) -> upload it to the azure storage.

In the current state of how this tool works, writing files to in-memory and then zip and export to storage is 1 unit of work. Currently this is what this tool does and code reflects that.

* Why are we moving away from this? 
* Why we are making this configurable? (Are we doing all this to fit any scenario which we don't know about?) 

Unless things are clear why are we adding new configurable flags, like for example only thing configurable before was 1-thing which is `storage account` name and key. Now we have so many moving parts.

Thanks,

davidkydd: Both zipping and non-zipping have their advantages and disadvantages, therefore we should support both and expose the configuration necessary for this:

- We know we need a LocalMachineExporter, which should use the ZIP function in order to reduce the amount of data that needs to be sent over the network to the CLI.

- We know we need an AzureBlobExporter, this should probably NOT use the ZIP function because we don't want to increase the amount of processing cycles which periscope consumes on the customer cluster, in the case that it is only outputting to AzureBlob storage and the bits are staying within the microsoft network.

Therefore we know we need to support both ZIP and non-ZIP. 

More generally - declarative config should not be considered as moving parts, it is a flat statement of what we want the tool to do, the moving parts are on the inside baked into the image. Adding additional config isn't something to fear or shy away from - we should look to make periscope configurable and tunable via exposing config values where ever it makes sense, as this provides a great deal of flexibility to adapt or to support new scenarios which we might not have considered. 

If we don't add new configuration flags, then the config is still there - it is just set to a single value and hard-coded into the image.

cmd/aks-periscope/aks-periscope.go
arnaud-tincelin: My position is as follow:
- expoters shall be variabilized as they depend on the ecosystem around periscope
- diagnosers and collectors don't have the same kind of dependencies and shall not be variabilized. Periscope provides a bunch of logs, it's up to the user to use them or not, to filter them, sort them etc ... later on
- I see no issue with backcompat, these are env variables. If the code don't use them it won't break anything. Why the need to have these vars anyway?

pkg/exporter/azureblob_exporter.go
arnaud-tincelin: There is no need to select exporters for now anyway so I believe this is just creating noise. I'd rather come up with an elegant solution the day we need to do this.

cmd/aks-periscope/aks-periscope.go
arnaud-tincelin: Then take zip code and put it in the localmachineexporter and remove it from there. 
You were speaking about reducing the amount of data sent over the network; today because we have this zip at this location, we are sending a lot more than necessary (because others diagnoser / collectors are already sending the same data)

pkg/exporter/azureblob_exporter.go
davidkydd: well - the day we need to do this is today :) this PR would have included the motivation for doing this (the localMachineExporter) but on request I split the original PR up into multiple smaller PR, and haven't submitted the localMachineExporter one yet as I only want to have to rebase in one direction.

How is it inelegant to be doing this in the same way as collectors and diagnosers, and without having to introduce any new concepts? The pre-existing COLLECTOR_LIST is an inelegant solution by comparison as it introduces another layer of indirection in the middle

cmd/aks-periscope/aks-periscope.go
Tatsinnit: 💡Also,  Why do we need `LocalMachineExporter` ? what benchmarking do we have for the amount of processing cycle it uses right now and improvement this will do? re:  `ZIP function because we don't want to increase the amount of processing cycles which periscope consumes on the customer cluster,` 

There are many open ended questions again, like what all goals this PR eventually have.

We handle the in-memory logs and what current mechanism is to make less noise and in the end zip and export, this `LocalMachineExporter` what is the use case for this? Like in what scenario this will happen. Currently this tool don't support it at-all because the pods in each nodes upload these logs.

davidkydd: why restrict zip functionality to localMachineExporter? It is a general feature and other exporters in future might wish to make use of it - even the AzureBlobExporter.

Yes we are sending more data than necessary currently - with this config flag other tools / consumers of periscope can choose to disable the zip feature if they wish.

Also note this PR isn't aimed at and doesn't change the current behaviour of the zip (the flag value defaults to true)

cmd/aks-periscope/aks-periscope.go
davidkydd: We are going around in circles... Arc scenarios need the ability to disable some of the current collectors. Back compat is for the existing CLI implementation which might make use of COLLECTOR_LIST already (Tats can confirm)

cmd/aks-periscope/aks-periscope.go
davidkydd: Arc need LocalMachineExporter for the case where a cluster does not have access to egress to Azure storage due to firewall or other egress restrictions.

We don't need to benchmark - it is a general statement that zip requires CPU and reduces file size

Tatsinnit: 💡 Cool, so this PR #57 is somehow connected to this work, i.e. this is ground up for for achieving localMachineExporter? Because if it is for localMachineExporter we should give some thought to the actual use of local node file save vs where_to_in_local_save?

* suggestion: I would recommend to open a WorkItem clearly summarise or detail the purpose of these refactor, probably at this repo level for this and possibly what the end goal for refactor, or this is all-in-one refactor PR?

* Regarding data and other discussion, I would recommend thinking of `client-go` adaptation. I might land something as first adaptation point but any other PR should first think of how we can use `client-go` for various other PR(s) and is it feasible, (or need discussion etc)

Thanks 🙏

davidkydd: Not PR #57 exactly but yes - I have an equivalent LocalMachineExporter which is rebased on the latest changes that have since gone into master.

This PR does not change the location of the node file save, if we want to debate or change this we can do this later - outside of this PR.

This is a refactor to enable the current set of requirements which we have discussed many times - I don't see the value in opening new workitems for this work that is already completed?

PR #85, opened 2021-07-20
Singleton implementation for GetHostName.

Tatsinnit:
This PR intends to Fix the multiple `GetHostName` calls to Singleton call using `Sync.Once.Do`

This PR also try to attempt and simplify the deep parameter passing mechanism. Why using `Sync.Once` because its a thread safe go existing pkg funct which we can use.  https://pkg.go.dev/sync#Once .

Please note: feel free to add your variable naming preferences but this way will ensure that the call for the actual `run command` only happens once. **Advantage** of this approach is that all changes reside under `utils` umbrella, no extra func parameter will be required, and it will guarantee the actual code is ran only once. 

Another, pro is that the call to the method stays same, under the hood it only gets called once. For future any new diagnostics or anything which needs to call the `utils.GetHostName` don't need to worry about running many times, and no more parameter passing as well.

* **Tested with Image local**: docker.io/tatsat/singleton-call-changes 

This pR also simplifies this PR: https://github.com/Azure/aks-periscope/pull/82 

Thanks guys. 🙏

Screen Shot 2021-07-20 at 9 11 12 PM

**Unit test result**

Screen Shot 2021-07-21 at 2 06 46 AM

### DIFF
--- a/pkg/utils/helper.go
+++ b/pkg/utils/helper.go
@@ -24,6 +24,8 @@ const (
 	AzureStackCloudName = "AzureStackCloud"
 )
 
+var GetHostNameFunc = GetHostNameSingleton()
+
 // Azure defines Azure configuration
 type Azure struct {
 	Cloud string `json:"cloud"`
@@ -95,8 +97,8 @@ func GetStorageEndpointSuffix() string {
 }
 
 type HostName struct {
-	hostName string
-	err      error
+	HostName string
+	Err      error
 }
 
 var singletonHostName *HostName
@@ -112,8 +114,8 @@ func GetHostNameSingleton() *HostName {
 		}
 
 		singletonHostName = &HostName{
-			hostName: hostname,
-			err:      err,
+			HostName: hostname,
+			Err:      err,
 		}
 	})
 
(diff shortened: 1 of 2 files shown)


davidkydd (dismissed):
Looks good, maybe add a unit test or two?

davidkydd:
Sorry for premature approval - forgot about testability, this will maintain current tight coupling which we should move away from, and would make it impossible to inject a value of hostname for unit testing those components which consume it. We should either pass the value as a parameter or abstract the singleton behind an interface.

Tatsinnit:
> Sorry for premature approval - forgot about testability, this will maintain current tight coupling which we should move away from, and would make it impossible to inject a value of hostname for unit testing those components which consume it. We should either pass the value as a parameter or abstract the singleton behind an interface.

Tight coupling as to unit testing? both can be go mocked. I am adding test to this and you can see how gomock can be used along with Gomega which will provide better verb. Or are you referring to something else?

davidkydd (approved):
Callers are still tightly coupled to this implementation but think I'm not explaining it properly, and it would be a premature optimization anyway so this is perfect for now. Thanks!

PR #86, opened 2021-07-20
fix CRD bug

Tatsinnit:
This PR captures fix under this PR: https://github.com/Azure/aks-periscope/pull/82 for the CRD fix. (in more succinct form and the refactor to use singleton is done using sync.once.do in PR #85 )

Thanks.,

cc: @arnaud-tincelin , @davidkydd

### DIFF
--- a/pkg/utils/helper.go
+++ b/pkg/utils/helper.go
@@ -251,7 +251,7 @@ func GetCreationTimeStamp() (string, error) {
 }
 
 // WriteToCRD writes diagnostic data to CRD
-func WriteToCRD(fileName string, key string) error {
+func WriteToCRD(fileContent string, key string) error {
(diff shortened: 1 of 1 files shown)


pkg/utils/helper.go
```diff
@@ -251,20 +251,15 @@ func GetCreationTimeStamp() (string, error) {
 }
 
 // WriteToCRD writes diagnostic data to CRD
-func WriteToCRD(fileName string, key string) error {
+func WriteToCRD(fileContent string, key string) error {
```
davidkydd, line 254: Can we rename the parameter to crdContent or similar? This doesn't represent a file anymore

davidkydd (changes requested):

davidkydd (approved):
👍

PR #87, opened 2021-07-21
support for kind clusters

davidkydd:
storageContainer name generation logic changes
new containerd container log collector
utils.GetApiServerFQDN changes to read from alternative conf location


Tatsinnit:
👍 Thank you, will take a look at this by tonight, also for reference \ gentle ping to @songjiaxun and @shwetamurali for the `containerDlog` specific comments, further to archive it for the readers, and more details reside under this PR as well : https://github.com/Azure/aks-periscope/pull/73 

: ) side note: the title might need changing as this PR is not for `Kind cluster support` but to **enable** `Conatianer Logs collector`

pkg/exporter/azureblob_exporter.go
```diff
@@ -76,6 +76,24 @@ func createContainerURL() (azblob.ContainerURL, error) {
...
+// GetStorageContainerName get storage container name
+func getStorageContainerName(APIServerFQDN string) (string, error) {
+	containerName := strings.Replace(APIServerFQDN, ".", "-", -1)
+
+	//TODO DK: I really dont like the line below, it makes for weird behaviour if e.g. .hcp. or -hcp- is in the fqdn for some reason other than being auto-added by AKS
+	length := strings.Index(containerName, "-hcp-")
+
+	if length == -1 {
+		maxLength := len(containerName)
+		length = int(math.Min(float64(maxLength), float64(maxContainerNameLength)))
```
Tatsinnit, line 88: 💡 What is this logic? not able to follow this. why const is `float64` ? Why are we finding min at this point is not clear either, (Is this replacement of any old logic if you can point me to please)

Tatsinnit (changes requested):
💡 Thanks for PR, other folks might have more comments, here are my initial comments, this PR is cumulation of 2 things - supporting the `kubeadm` location of config file and also the `containerd log`.

Below are some recommendations. Thank you 🙏

tests/helper_test.go
```diff
@@ -50,3 +50,64 @@ func TestGetHostNameFailureCase(t *testing.T) {
 	g.Expect(err).To(BeElementOf(errors.New("Fail to get host name: an error")))
```
davidkydd, line 50: is there a reason this file was added to separate "Tests" folder? Usual go project convention is that /tests/ contains external or integration type tests, and unit tests exist in the folder alongside the code-under-test. Can I move this file into /pkg/utils/ ?

pkg/utils/helper.go
```diff
@@ -132,28 +135,57 @@ func GetHostName() (string, error) {
...
+
+			return host, nil
 		}
 	}
-
 	return "", errors.New("Could not find server definitions in kubeconfig")
 }
 
+//ReadKubeletConfig reads the kubeletConfig from the node
+func ReadKubeletConfig() (string, error) {
```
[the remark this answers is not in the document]
sophsoph321: Hi @Tatsinnit, I have not tested periscope against GKE or EKS, so am not aware of the kubeconfig file locations for these distros.

pkg/exporter/azureblob_exporter.go
davidkydd: this replaces the code in lines 41-46 of master. 

No function signature overloading in Go means math.Min is only defined for float64 and we have to pollute with casts like this. Alternative is probably something like adding an explicit "MinInt" func.

We need to get the min to enforce 63 char limit, rest of the logic does stuff needed to make the test cases pass :)

At some point its worth refactoring the -hcp- check to give less "false positives" by pattern matching only to strings ending with "hcp.<region>.azmk8s.io"

pkg/utils/helper.go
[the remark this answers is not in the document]
davidkydd: Agree - pulling out checkIfKubeConfigFileExist as a separate func is better. Can't use os.stat because this has to run on the Host - can be refactored to use a volume mount later.

pkg/utils/helper.go
```diff
@@ -132,28 +135,57 @@ func GetHostName() (string, error) {
...
+	return output, nil
+}
+
+// GetAPIServerFQDN gets the API Server FQDN from the kubeconfig file
+func GetAPIServerFQDN() (string, error) {
+	output, err := ReadKubeletConfig()
+	if err != nil {
+		return "", err
+	}
+	fqdn, err := ParseAPIServerFQDNFromKubeConfig(output)
```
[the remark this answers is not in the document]
davidkydd: agree - this should be private. Moving helper_test to same folder so it can still call it

pkg/collector/containerlogs_collector_containerd_test.go
```diff
@@ -0,0 +1,198 @@
+package collector
```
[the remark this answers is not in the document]
davidkydd: done

pkg/collector/containerlogs_collector_containerd.go
```diff
@@ -0,0 +1,160 @@
+package collector
```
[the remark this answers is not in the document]
davidkydd: done

tests/helper_test.go
davidkydd: moved so we can call private methods

pkg/utils/helper.go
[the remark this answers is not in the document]
davidkydd: refactor to pull out func and use const's done @Tatsinnit 

Actually /etc/kubernetes/kubelet.conf might be the standard location as I think all the other distros (ARO, AzureStack etc) have it located here and AKS is the odd-one-out.

tests/helper_test.go
```diff
@@ -1,52 +0,0 @@
-package main
```
Tatsinnit, line 1: 💡 Please revert this to the location under one testSuite: i.e. `tests` folder will contain all the tests: 

The whole point of unit tests is to test the "external interface" (ie public functions) to your package. That is unit tests should always be "white box" tests (see White Box Testing). That way you can refactor your code and your tests won't break.

I think this is moved to the folder where all collector so that you have access to a private method which is not correct.

pkg/collector/containerlogs_containerd_collector_test.go
```diff
@@ -0,0 +1,198 @@
+package collector
```
Tatsinnit, line 1: 💡 Test should reside in `tests` folder, also the whole point of unit tests is to test the "external interface" (ie public functions) to your package. That is unit tests should always be "white box" tests (see White Box Testing). That way you can refactor your code and your tests won't break.

Moving this to `tests` folder not only makes it structured as well.

pkg/utils/helper.go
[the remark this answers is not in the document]
Tatsinnit: 💡 I see that basically the whole location of test files are moved from location to just test `private-method` instead like mentioned above test the `GetAPIServerFQDN()`, keep the test suite location as `tests` folder where all the tests will reside, but test the `public-scope methods`.

pkg/collector/containerlogs_containerd_collector_test.go
Tatsinnit: 💡 **Possible solution\proposal:** 

The way these new `private scooped functs` are written seems like they are supporting `helper` to the specific package, so if we really wan to to test these private functions, we can do so by structure and scope change.

* Utils -- is a place of utility pkgs - so make that `containers-helper.go` under `utils` that way all these private functs which needs to be unit-tested will become `public` scope and as a separate unit.
* under `tests` folder there will be tests for these new `specific-container_helper_test.go` which can test these methods.
* Also it will be good idea to test `Collector` but I would recommend only write test for them if there is genuinely we can test anything, i.e. mocking something will not help if it by passes bunch of stuff.

pkg/utils/helper.go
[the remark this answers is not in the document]
davidkydd: Its pretty standard golang project structure to put unit test files in the same folder the files they are testing. All AKS projects I have seen are like this. Any particular reason you want to keep tests in a separate folder?

I'm not convinced there is never a case for unit testing private (internal) methods. In this case I think it is the appropriate choice - otherwise we will have to add significant code to mock out and or/ redesign the internal representation just to maintain the current level of testing and with no functional improvements. The language explicitly supports this capability so we should use it where appropriate.

Furthermore this code will be significantly refactored very soon when we change it to use a volume mount rather than directly exec'ing commands on the host. We shouldn't overinvest in its current form unless the benefits are obvious. What are the benefits of only testing the public API here?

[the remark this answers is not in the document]
Tatsinnit: 💡 I think this might help in [better summarizing the point of view which I am coming from](https://softwareengineering.stackexchange.com/questions/100959/how-do-you-unit-test-private-methods). Unless your private method is so complex that it needs its own unit-test. WE can totally rationalize that in here and move forward. What do you think?

>>You generally don't unit test private methods directly. Since they are private, consider them an implementation detail. Nobody is ever going to call one of them and expect it to work a particular way.

>> You should instead test your public interface. If the methods that call your private methods are working as you expect, you then assume by extension that your private methods are working correctly.

pkg/collector/containerlogs_containerd_collector_test.go
Tatsinnit: 💡 I am fine with location et. al. but I think you just need to rationalize for these tests for Private methods, in case you absolutely fine with it, I am happy to +1 this, but why not add test for public method? Or is it something you tried already? (I do know bunch of these legacy collectors are hard to test)

pkg/utils/helper.go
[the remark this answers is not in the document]
davidkydd: I think this is complicated enough that  it's worth unit testing the internal method in isolation. If we didn't know it was going to be refactored soon then I would consider pulling it out into a separate package or something, but don't want to over invest.

Also note the closest parallel in Java is "no access modifier defined" and for C# is "internal" keyword. Not "private".

So are we good for now to leave this as is?

pkg/collector/containerlogs_containerd_collector_test.go
davidkydd: Sorry I missed this earlier - I think it is incorrect to consider these as class private methods for the purpose of whether to unit test them - they are package internal, or "non-exported" methods accessible anywhere within the "collector" package and from what I have seen it is conventional golang to unit test them like this.

Agree that we should look to add some general tests for "Collect" at the public interface level too, and agree with the caveat to be careful of mocking out the useful parts.

I had a look at adding equivalent tests for the exported (public) method instead but the amount of additional complexity or mocking required to get the same level of test coverage as we have currently put me off doing this. Again this collector will be being refactored soon to use a volume mount and potentially a different internal structure due to needing to target pod logs, so I didn't want to over commit.

Potentially the concept of a "selector" is common enough across a range of collectors that we could pull it out into being its own type, but can look at that later too.

pkg/utils/helper.go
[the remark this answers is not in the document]
Tatsinnit: Sounds good 👍

Tatsinnit (dismissed):
👍 I have tested it locally under this image `docker.io/tatsat/kubekind-container-d` here is +1 for approval, Arnaud is back and will eye ball this sometime today, can we please hold-off merging late evening and then we can merge it early tomorrow or something. cc: @arnaud-tincelin , thanks

Please feel free to re-test for double assurance with different image.

Thanks,

pkg/collector/containerlogs_containerd_collector.go
```diff
@@ -0,0 +1,160 @@
...
+			if doesSelectorSelectContainerLog(containerLog, selector) {
+				selectedContainerLogs = append(selectedContainerLogs, containerLog)
+			}
+		}
+	}
+	return selectedContainerLogs
+}
+
+func doesSelectorSelectContainerLog(containerLog ContainerLog, selector ContainerLogSelector) bool {
+	if selector.namespace == allNamespacesSelector {
```
arnaud-tincelin, line 91: why don't u put all conditions in one if?

return selector.namespace == allNamespacesSelector && containerLog.namespace == selector.namespace && strings.HasPrefix(containerLog.containerName, selector.containerNamePrefix)

pkg/collector/containerlogs_containerd_collector.go
```diff
@@ -0,0 +1,160 @@
...
+// where I assume # is incremented and a new log created for each container restart (starts at # == 0)
+const containerLogDirectory = "/var/log/containers"
+
+func (collector *ContainerLogsContainerDCollector) GetData() map[string]string {
+	return collector.data
+}
+
+// Collect implements the interface method
+func (collector *ContainerLogsContainerDCollector) Collect() error {
+	selectorStrings := strings.Fields(os.Getenv("DIAGNOSTIC_CONTAINERLOGS_LIST"))
```
arnaud-tincelin, line 53: why this env variable?

I think any collector should collect all possible data.
What is done with (filtering, sorting, ...) might not be aks-periscope scope

pkg/collector/containerlogs_containerd_collector.go
```diff
@@ -0,0 +1,160 @@
...
+			containeruid:  containerNameWithIDSplitOnDash[indexOfUid],
+			filepath:      path.Join(directoryPath, logFile),
+		})
+	}
+
+	return containerLogs
+}
+
+func getAllContainerLogFilesThatHaveEverRunOnHost() ([]string, error) {
+	output, err := utils.RunCommandOnHost("ls", containerLogDirectory)
```
arnaud-tincelin, line 151: As you still use `RunCommandOnHost` I suppose we don't want to get rid of it?
Keep it mind that this implies no possible support for other OS, bad error handling, no complex structure (since we can only retrieve strings), no streaming, etc ...

arnaud-tincelin (changes requested):

pkg/collector/containerlogs_containerd_collector.go
davidkydd: we do want to get rid of RunCommandOnHost (or at least minimize its usage for all the good points which you bring up - I think we will want to retain the capability unless it becomes the only blocker to scoping down the permissions to e.g. allow periscope to run as non-privileged pod).

This can be refactored to use volume mount, but this required significant changes to the internal structures so I punted on it for me

pkg/collector/containerlogs_containerd_collector.go
davidkydd: I disagree, collectors should not collect all possible data (although they should be able to). This was one of the original periscope requirements and AKS leverages this to provide a guarantee that when support runs periscope and examines the outputs in the process of working on a support case, we won't e.g. collect and examine data from customer namespaces which are running their workloads, and which might contain sensitive data

pkg/collector/containerlogs_containerd_collector.go
davidkydd: The line you pasted would change the functionality, an equivalent single line would be:
return (selector.namespace == allNamespacesSelector) || (containerLog.namespace == selector.namespace && strings.HasPrefix(containerLog.containerName, selector.containerNamePrefix))

But the current form keeps the "special behaviour" as a separate case and is more readable

pkg/collector/containerlogs_containerd_collector.go
Tatsinnit: 💡Thoughts:  Might be good if shared a `os.Getenv("DIAGNOSTIC_CONTAINERLOGS_LIST")` input will look like? , and to make things bit more clear. 

* What current impact this has? (I think all this is new build only specific to `container-d` work I reckon? which is separate functionality altogether or is this impacting the current container log mechanism) 
* What is the current behaviour this will effect/affect if-at-all it will? 
* What feature implication / impact will it have in existing use-case. (I think none, please do correct me)

Thanks. (above might help to see wider scope)

davidkydd: this is needed to let this new collector match the existing behavior for the existing containerlogs collector (which this containerd collector is just a different implementation of). 

Examples of where it is defined:

https://github.com/Azure/aks-periscope/blob/2ea86f05fcfab1a77ea7f553e8801428633ff04c/deployment/aks-periscope.yaml#L127

https://github.com/Azure/aks-periscope/blob/2ea86f05fcfab1a77ea7f553e8801428633ff04c/deployment/config-map.yaml#L6

and of where it is referenced and used in the original docker based containerlog collector:

https://github.com/Azure/aks-periscope/blob/2ea86f05fcfab1a77ea7f553e8801428633ff04c/pkg/collector/containerlogs_collector.go#L28

sophsoph321:
Hi, Thank you @Tatsinnit for providing a test image. Thank you @arnaud-tincelin for comments, and thank you @davidkydd for making these changes. Just wanted to check in as to what is the ETA that this PR will be merged. Is there anything that you need from the arc team in order to close on this PR? 
Adding @safeermohammed for visibility

davidkydd:
> Hi, Thank you @Tatsinnit for providing a test image. Thank you @arnaud-tincelin for comments, and thank you @davidkydd for making these changes. Just wanted to check in as to what is the ETA that this PR will be merged. Is there anything that you need from the arc team in order to close on this PR?
> Adding @safeermohammed for visibility

Hi Sophie,
I should be able to get the changes which have been asked of this PR into review by EOD today or tomorrow at latest (am on call today), hopeful to be merged by end of the week. No questions for you yet but will reach out if we have any 🙂. Apologies for delays.

PR #93, opened 2021-08-04
Updated pipeline

arnaud-tincelin:
Reference is that it uses the Kind cluster, with test infra, with three builds for **Lint build and test,** provides structured infrastructure to the test.

rest details are more clear in the yaml file.

Thanks.


Tatsinnit (approved):
👍 Looks good, thank you so much for structured approach. 🙏

PR #95, opened 2021-08-06
Use os.open function for reading files.

Tatsinnit:
This PR kind of closes all the discussion where it was recommended not to use `RunCommandOnHost` for things like `cat` for file because that can be easily done at via `os.Open` which open files. Much better and reduces over-use of `RunCommandonHost`.

* Test with this image: `docker.io/tatsat/redundant-runcommand-use` 

Thanks.

### DIFF
--- a/pkg/collector/dns_collector.go
+++ b/pkg/collector/dns_collector.go
@@ -24,7 +24,7 @@ func (collector *DNSCollector) GetName() string {
 func (collector *DNSCollector) Collect() error {
 	result := make(map[string]string)
 
-	output, err := utils.RunCommandOnHost("cat", "/etc/resolv.conf")
+	output, err := utils.ReadFileContent("/etc/resolv.conf")
 	if err != nil {
 		return err
 	}
--- a/pkg/collector/nodelogs_collector.go
+++ b/pkg/collector/nodelogs_collector.go
@@ -1,9 +1,10 @@
 package collector
 
 import (
-	"io/ioutil"
 	"os"
 	"strings"
+
+	"github.com/Azure/aks-periscope/pkg/utils"
 )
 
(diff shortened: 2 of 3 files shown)


pkg/collector/dns_collector.go
```diff
@@ -24,7 +24,7 @@ func (collector *DNSCollector) GetName() string {
 func (collector *DNSCollector) Collect() error {
 	result := make(map[string]string)
 
-	output, err := utils.RunCommandOnHost("cat", "/etc/resolv.conf")
+	output, err := utils.ReadFileContent("/etc/resolv.conf")
```
arnaud-tincelin, line 27: The previous command was running on the host, this one will run on the container. Is it working as expected or should we have a mount to do that?

pkg/collector/nodelogs_collector.go
```diff
@@ -29,13 +29,12 @@ func (collector *NodeLogsCollector) Collect() error {
 
 	for _, nodeLog := range nodeLogs {
 
-		output, err := utils.RunCommandOnHost("cat", nodeLog)
+		output, err := utils.ReadFileContent(nodeLog)
```
arnaud-tincelin, line 32: same here

arnaud-tincelin (changes requested):

pkg/collector/dns_collector.go
Tatsinnit: Yeah, so they are already there: `docker.io/tatsat/redundant-runcommand-use` image shows the correct log.

Tatsinnit:
💡 FYI: closing this in favour for this PR: https://github.com/Azure/aks-periscope/pull/109 - it makes sense to do this individually at the each collector level when adding tests. Which will lead to better progress. Thanks heaps.

PR #96, opened 2021-08-07
Fix: remove hardcoded path to kubeconfig file

arnaud-tincelin:


safeermohammed (approved):
:shipit:

pkg/collector/networkoutbound_collector.go
```diff
@@ -57,12 +50,6 @@ func (collector *NetworkOutboundCollector) Collect() error {
 			URL:  "kubernetes.default.svc.cluster.local:443",
 		},
 	)
-	outboundTypes = append(outboundTypes,
```
Tatsinnit, line 60: 🐛 Can we not make this removal part of this PR please, we don't know why it existed, @rzhang628 , @qpetraroia or @palma21 - If anyone of you have any insight for this please? removing this means taking small functionality out of periscope and since its in `Azure` org I would prefer wider eyes on this please. Thanks

pkg/exporter/azureblob_exporter.go
```diff
@@ -33,23 +29,18 @@ func NewAzureBlobExporter(creationTime, hostname string) *AzureBlobExporter {
...
-	APIServerFQDN, err := utils.GetAPIServerFQDN()
-	if err != nil {
-		return azblob.ContainerURL{}, err
-	}
+	accountName := os.Getenv("AZURE_BLOB_ACCOUNT_NAME")
+	sasKey := os.Getenv("AZURE_BLOB_SAS_KEY")
+	containerName := os.Getenv("AZURE_BLOB_CONTAINER_NAME")
 
-	containerName := strings.Replace(APIServerFQDN, ".", "-", -1)
-	containerLen := strings.Index(containerName, "-hcp-")
```
Tatsinnit, line 42: 💡 This logic needs to be extracted in some kind of git issue, (cc: @sophsoph321  - because `az-cli` will now do this before supplying the container name as `APIServerName` ) also fyi: @rzhang628 and @qpetraroia  - `vscode` also need similar tracking workitem in its queue and I can work on that change.

pkg/exporter/azureblob_exporter.go
```diff
@@ -33,23 +29,18 @@ func NewAzureBlobExporter(creationTime, hostname string) *AzureBlobExporter {
...
-	if containerLen == -1 {
-		containerLen = maxContainerNameLength
+	if accountName == "" || sasKey == "" || containerName == "" {
```
Tatsinnit, line 36: 🐛 Please add check for `containerName` length. `3 to 63` is the min and max - otherwise it will fail and in that case how will we handle that scenario? https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names 

There are bunch of stuff removed and will sit under consuming tool which looks cool.

pkg/utils/helper.go
```diff
@@ -132,28 +131,6 @@ func GetHostName() (string, error) {
 	return hostName.HostName, nil
 }
 
-// GetAPIServerFQDN gets the API Server FQDN from the kubeconfig file
-func GetAPIServerFQDN() (string, error) {
```
Tatsinnit, line 136: 💡 Same as above comment, that we could move this to `netwrokbound_collector.go` but completely removing which include taking small chunk of log out of the logs which gets dumb - looks ok to my eyes but need folks like @JunSun17 or @rzhang628, @qpetraroia to verify what are the key reasons it was there at first place.

docs/kustomize.md
```diff
@@ -1,50 +0,0 @@
-# Deploy with Kustomize
```
Tatsinnit, line 1: 💡 Can we please keep this as a document, will be good to update the instructions for newbieusers who want to see how `kustomize` file will fit together and how they can consume this in their tool or their purpose. Thanks

cc: @rzhang628 and @qpetraroia

Tatsinnit (changes requested):
💡 Thank you so much for this 🙏 +1 for This PR as it simplifies the use of `containerName` and various things it unblocks in future. Issue was there because from day-1 the `APIServerName` was used as `containerName` : detailed in #99 

This will sit nicely and fix any of the ARC scenario as well.

Few things to stock take please:

* Can we please open connecting issues to respective repositories for this work please as communication: https://github.com/Azure/azure-cli-extensions/ and https://github.com/Azure/vscode-aks-tools - Please feel free to assign vscode issue to me. ( @rzhang628  and @qpetraroia ) also @sophsoph321 - if you are willing to help this is where you could help in putting and PR for the `az-cli` fix.
* There are couple of removal of small `AKS tunnel` logs with this PR does either we can remove it completely or add we need to discuss why it existed and make informed decision.

I do know that for any breaking changes we just need plan and some part of above comment is circling around that preparation. (we need to give enough time for other tools to adapt these changes)

* **Important** and might be worth a discussion : I am happy to merge once we catered some of thee recommendations and keep these changes as release `v0.5` or something, but we just need to think of scenario if there is something we need to fix for any other consuming tool where we will end with this change which needs work at that consuming tool.  **to**: @rzhang628 , @qpetraroia , @palma21 , @sophsoph321 - as fyi and to please stock take and we can start the work in the consuming tool on the side to avoid any convoluted situation. @JunSun17 - if you have any opinions. Thanks you all.

**Rest of PR looks great**, but I think a plan is key for all including PMs 👍 thanks heaps

deployment/examples/kustomization_storage_account.yaml
```diff
@@ -1,14 +1,19 @@
+# This is an example file to use an Azure Storage Account to export data
```
Tatsinnit, line 1: 👍 Thank you for this, to have a sample file for consuming tools as example or sample set - I fee like this could directly be used as well. (will check it out)

pkg/collector/networkoutbound_collector.go
arnaud-tincelin: what should we set instead?

pkg/exporter/azureblob_exporter.go
arnaud-tincelin: yes, this needs to be documented in the release 0.5. Not sure what would be the benefit of an issue

pkg/exporter/azureblob_exporter.go
arnaud-tincelin: hmmm sure but in the same logic we could also check the accountname format, the saskey format etc ... I'd rather not do checks here. Anyway the user will eventually have an error from the service

docs/kustomize.md
arnaud-tincelin: documentation is in the example file we added. This brings no value so we should remove it

pkg/exporter/azureblob_exporter.go
Tatsinnit: Cool, so the reason we need this is because periscope as a tool from beginning  has working contract with “az-cli” via kollect command and latter vscode started using so anything effecting long term to these tools or any change of contract needs to be highlighted.

with tracking workitem we need to inform those consuming tools: (give enough information for this change) 

refrence:

* https://github.com/Azure/aks-periscope#dependent-consuming-tools-and-working-contract
* https://github.com/Azure/aks-periscope#user-guide

thanks 🙏

pkg/exporter/azureblob_exporter.go
Tatsinnit: 💡This logic was already pre-existing in this tool, if we keep that length as it is it and do not remove we take pain from other consuming tool managing their own version of this?

pkg/exporter/azureblob_exporter.go
arnaud-tincelin: I think these 2 paragraphs have to be removed from our README:
- first one: we are not developping aks-periscope only for vscode & az cli; some others might use it and anyway noone can change this file unless a PR is approved. + we already documented this file as deprecated
- 2nd one: the detail of how to use an extension of az-cli shall be in the extension repo. We are not going to watch the extension repo so that we can align our README. However we can indicate that az-cli can be used to deploy aks-periscope and add a link to our repo to the extension

Tatsinnit:
Thanks Arnaud, @sophsoph321 can you please test this as a test image? cc: @safeermohammed  Thank you.

arnaud-tincelin:
> Thanks Arnaud, @sophsoph321 can you please test this as a test image? cc: @safeermohammed Thank you.
> 
> * can you publish your local image or you need hand for that please? (Don't hesitate?) thank you

Feel free to use ghcr.io/arnaud-tincelin/aks-periscope:0.5

deployment/examples/kustomization_storage_account.yaml
```diff
@@ -17,10 +22,12 @@ patches:
...
   patch: |-
     - op: add
-      path: '/spec/template/spec/containers/0/env/-'
+      path: '/spec/template/spec/containers/0/env'
       value:
-        name: AZURE_BLOB_ACCOUNT_NAME
-        value: <name>
+        - name: AZURE_BLOB_ACCOUNT_NAME
+          value: <name>
+        - name: AZURE_BLOB_CONTAINER_NAME
```
sophsoph321, line 29: Currently our CLI code calls the aks-periscope.yaml file https://github.com/alphaWizard/azure-cli-extensions/blob/5dd22cbf7cf03e26f374a62504c90a5403bdd1c7/src/connectedk8s/azext_connectedk8s/_utils.py#L757. To avoid a large refactoring of the CLI code, is it possible add the AZURE_BLOB_CONTAINER_NAME field to the single yaml file as well?

sophsoph321 (changes requested):

Tatsinnit:
> Thank you @arnaud-tincelin for helping us get unblocked. I have tested using the image you provided on the different distros for our MVP and those changes look good.
> If we can add the env variable AZURE_BLOB_CONTAINER_NAME in the aks-periscope.yaml file, we should be good to merge this PR.

Thanks @sophsoph321 for checking, that is good that image is good for your scenarios. - To expedite this, I would recommend - share your `*.yaml` you are using and we could keep it under examples folder under `deployment` which will be only short term solution until you start to do whole `kubectl kustomize` at your functionality level. Just as a thought: We **cannot** change `aks-periscope.yaml` because it is used by  because it is consumed `az-cli` and `vscode` - Please read this tracking item - https://github.com/Azure/aks-periscope/issues/99 

Thanks.

sophsoph321:
> Thanks @sophsoph321 for checking, that is good that image is good for your scenarios. - To expedite this, I would recommend - share your `*.yaml` you are using and we could keep it under **examples folder** under `deployment` which will be only short term solution until you start to do whole `kubectl kustomize` at your functionality level. Just as a thought: We **cannot** change `aks-periscope.yaml` because it is used by because it is consumed `az-cli` and `vscode` - Please read this tracking item - #99
> 
> And, If I may suggest, lets add a sample way how to use `kubectl kustomize` command, and to build it with storage_details.
> 
> Thanks.

@arnaud-tincelin Spoke to @Tatsinnit offline. We used the "kubectl kustomize" to construct a single yaml file for the kustomize template. The aks-periscope pods have a ContainerCreationError with the following error message: "Error: secret "azureblob-secret" not found". 
For the time being, I used this yaml file to test your code changes, the contents of this file can reside in the examples directory for kustomize (git will not allow me to attach yaml files, so it's in txt format)
[azure-k8s-periscope.txt](https://github.com/Azure/aks-periscope/files/6973164/azure-k8s-periscope.txt)

arnaud-tincelin:
Hi @sophsoph321 apologies, there were 2 mistakes in the example file:
- namespace was not specified so the secret was not created (add `namespace: aks-periscope`)
- value of the secret should not be base64 encoded (value of `AZURE_BLOB_SAS_KEY`)

Let me know if it works better this way

Here is the kustomization.yaml file you should end up with:
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
- https://github.com/Azure/aks-periscope/deployment//

namespace: aks-periscope

secretGenerator:
- name: azureblob-secret
  literals:
  - AZURE_BLOB_SAS_KEY=?sv=2020-08-04&ss=bfqt&srt=sco&sp=rwdlacuptfx&se=2021-08-15T16:02:03Z&st=2021-08-10T08:02:03Z&spr=https&sig=REDACTED
       
patches:
- target:
    group: apps
    kind: DaemonSet
    name: aks-periscope
    version: v1
  patch: |-
    - op: add
      path: '/spec/template/spec/containers/0/env'
      value:
        - name: AZURE_BLOB_ACCOUNT_NAME
          value: mystorageaccount
        - name: AZURE_BLOB_CONTAINER_NAME
          value: mycontainer
- target:
    group: apps
    kind: DaemonSet
    name: aks-periscope
    version: v1
  patch: |-
    - op: add
      path: '/spec/template/spec/containers/0/envFrom/-'
      value:
        secretRef:
          name: azureblob-secret

images:
  - name: aksrepos.azurecr.io/staging/aks-periscope
    newName: ghcr.io/arnaud-tincelin/aks-periscope
    newTag: '0.5'
```

sophsoph321:
> Hi @sophsoph321 apologies, there were 2 mistakes in the example file:
> 
> * namespace was not specified so the secret was not created (add `namespace: aks-periscope`)
> * value of the secret should not be base64 encoded (value of `AZURE_BLOB_SAS_KEY`)
> 
> Let me know if it works better this way

@arnaud-tincelin Thanks for looking into this. I have used your latest changes to generate a single yaml file via the following command: kubectl kustomize deployment >> test.yaml. And then from here, I do a kubectl apply -f test.yaml. The deployment was successful.
 
This approach however, may require a large refactoring of the CLI code for our MVP, which currently relies on the single yaml file.  We can consider this refactoring for a future release (a.k.a post-MVP) 
I can attach the test.yaml file that got generated, and we can add it to the example folder for kustomize, but it will not work for other users that want to use this yaml file for deployment, since the Secret section has metadata: azureblob-secret-6666jklm. This metadata will be different for everyone depending on the secret you pass in. If we change it to just azureblob-secret, the pods will end up with a containerCreationError. 

For a short-term solution, I'm thinking the quickest way, as Tats was suggesting, is to add the yaml file I shared above to your examples folder, will attach it here again.  Happy to sync up offline if you have a different solution. 
[azure-k8s-periscope.txt](https://github.com/Azure/aks-periscope/files/6977403/azure-k8s-periscope.txt)

sophsoph321 (approved):
Approving this PR. I added a workaround on my side for the CLI changes have tested regarding the AZURE_BLOB_CONTAINER_NAME field and the changes look good. So adding the yaml file I shared last night to the example folder is completely optional.

Tatsinnit (approved):
+1 Since @sophsoph321 approves it with scenarios which ARC needs to fit in, moving forward keeping this issue in mind: https://github.com/Azure/aks-periscope/issues/99 - we can approve this and move forward. 

Thanks you so much guys 🙏☕️❤️

PR #97, opened 2021-08-07
Remove docker collector

arnaud-tincelin:


Tatsinnit:
💡 This PR is way more simpler and looks good : - ) , **But** needs folks to make a informed decision for this hence @rzhang628 , @palma21, @justindavies or @qpetraroia  - please add some thoughts for this if you guys differ.

* **Docker deprecation here** : https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker/ 
* Question for PMs: Does that mean that we need workitem for Containerd-log-collector ?

Thanks heaps,
^Tats

Tatsinnit:
Ok, I am convinced this is no longer needed see below from latest default AKS cluster : 

Screen Shot 2021-08-12 at 9 38 15 AM

Tatsinnit (approved):
Docker is deprecated from Kubernetes 1.20 onwards. I have added the screenshot from a latest default cluster run.

PR #102, opened 2021-08-11
Updated readme

arnaud-tincelin:


README.md
```diff
@@ -35,124 +35,36 @@ AKS Periscope currently only work on Linux based agent nodes. Please use the scr
...
-2. Network configuration, includes Network Plugin, DNS, and Max Pods per Node settings.
+1. Network configuration, includes Network Plugin, DNS, and Max Pods per Node settings.
 
 ## User Guide
 
-AKS Periscope can be deployed by using Azure Command-Line tool (CLI). The steps are:
+AKS Periscope can be deployed by using Azure Command-Line tool (CLI). You will have to install the `aks-preview` extension to do so.  
+See [az cli aks-preview extension](https://github.com/Azure/azure-cli-extensions/tree/main/src/aks-preview/azext_aks_preview) for detailed information.
 
-0. If CLI extension aks-preview has been installed previously, uninstall it first.
```
Tatsinnit, line 57: 🐛 Why is this removed? Reason? (Like mentioned few times primary contract of this tool is to coexist with `az-cli` and `vscode` so lets sync up a meeting with @qpetraroia and @rzhang628 - but we cannot remove this, `Collect` command is one of the key use for this and supports few AKS scenarios.

Tatsinnit (changes requested):
There are few changes which takes the repo move away from early working contract and this PR removes some good documentation we need for those scenarios.

We need 

Like I mentioned in one of my comment: 

>>🐛 Why is this removed? Reason? (Like mentioned few times primary contract of this tool is to coexist with az-cli and vscode so lets sync up a meeting with @qpetraroia and @rzhang628 - but we cannot remove this, Collect command is one of the key use for this and supports few AKS scenarios.

README.md
arnaud-tincelin: yes but the kollect doc shall be in the kollect repo. This documentation should be moved. It's not aks-periscope reponsability to know how external tools use it.
Furthermore, to maintain this documentation up-to-date it would mean to watch for new releases of kollect command and update our readme accordingly - this is painful

Tatsinnit: 💡 I have blocked some time this week, might be there is missing information or could be background information, hence you probably are making these changes, the `collect` command was written at the same time and was the idea behind this tool. Hence in various places including the internals have updated a readMe file regarding consuming tools here: https://github.com/Azure/aks-periscope#dependent-consuming-tools-and-working-contract 

I would recommend that we can improve the documentation by proper - Wiki pages. Which might also include the consensus on other things. Removing the key usability docs from the documentation is a bad idea. We can discuss this in tomorrows catch-up (if possible) leaving this for wider PM visibility.

arnaud-tincelin: Don't you think it's weird to have the code in one repo and documentation in another one, managed by another team? That's why I suggested to put code & doc at the same place. If we agree on this PR all doc we remove from here, we put it in a PR on the kollect repo and add a link to this repo from our readme. In my opinion, that would make more sense.

Tatsinnit: Cool, @rzhang628 - has an idea for workitem to handle this in `wiki` for this repo. Which sounds like a great plan and thanks for that.

Tatsinnit:
closing this as the new read me changes reside here: https://github.com/Azure/aks-periscope/pull/186 ❤️☕️🙏

PR #114, opened 2021-09-09
Run tests from inside the cluster

arnaud-tincelin:


Tatsinnit:
💡 Regarding `codecov` decrease I think we should close this PR and re-open as separate PR: 

image

But Codecov is somehow stuck in random commit:

image

PR #117, opened 2021-09-15
dns logs will contain error rather than actual pod error visibility.

Tatsinnit:
This is the work for ARC for log level error visibility in case od DNS logs than pods level error visibility.

Thanks,

* Test image : `docker.io/tatsat/dns-logs-visibility` 

Thank you guys ☕️  : @arnaud-tincelin , @sophsoph321 and @safeermohammed


sophsoph321 (approved):
Verified with the private image and changes look good.

safeermohammed (approved):

arnaud-tincelin (approved):

PR #118, opened 2021-09-17
osm/smi: update dir structure and add test

SanyaKochhar:
- Updates to collect osm custom resource meshconfig per latest osm release
- Fixes collection of smi crds (rbac in cluster-role)
- Adds naming structure to filename: `meshName/resourceName_resourceType` and `smi/resourceName_resourceType>`
- Tested with kustomize `kubectl apply -k deployment`

### DIFF
--- a/deployment/cluster-role.yaml
+++ b/deployment/cluster-role.yaml
@@ -12,6 +12,3 @@ rules:
 - apiGroups: ["apiextensions.k8s.io"]
   resources: ["customresourcedefinitions"]
   verbs: ["get", "list", "watch"]
-- apiGroups: ["access.smi-spec.io", "specs.smi-spec.io", "split.smi-spec.io"]
-  resources: ["*"]
-  verbs: ["get", "list", "watch"]
--- a/pkg/collector/osm_collector.go
+++ b/pkg/collector/osm_collector.go
@@ -50,22 +50,22 @@ func (collector *OsmCollector) Collect() error {
 // callNamespaceCollectors calls functions to collect data for osm-controller namespace and namespaces monitored by a given mesh
 func (collector *OsmCollector) callNamespaceCollectors(monitoredNamespaces []string, controllerNamespaces []string, meshName string) {
 	for _, namespace := range monitoredNamespaces {
-		if err := collector.collectDataFromEnvoys(namespace); err != nil {
+		if err := collector.collectDataFromEnvoys(namespace, meshName); err != nil {
 			log.Printf("Failed to collect Envoy configs in OSM monitored namespace %s: %+v", namespace, err)
 		}
-		collector.collectNamespaceResources(namespace)
+		collector.collectNamespaceResources(namespace, meshName)
 	}
 	for _, namespace := range controllerNamespaces {
-		if err := collector.collectPodLogs(namespace); err != nil {
+		if err := collector.collectPodLogs(namespace, meshName); err != nil {
 			log.Printf("Failed to collect pod logs for controller namespace %s: %+v", namespace, err)
 		}
-		collector.collectNamespaceResources(namespace)
+		collector.collectNamespaceResources(namespace, meshName)
 	}
 }
 
 // collectNamespaceResources collects information about general resources in a given namespace
-func (collector *OsmCollector) collectNamespaceResources(namespace string) {
-	if err := collector.collectPodConfigs(namespace); err != nil {
+func (collector *OsmCollector) collectNamespaceResources(namespace string, meshName string) {
+	if err := collector.collectPodConfigs(namespace, meshName); err != nil {
 		log.Printf("Failed to collect pod configs for ns %s: %+v", namespace, err)
 	}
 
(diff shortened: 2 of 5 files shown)


deployment/aks-periscope.yaml
```diff
@@ -28,6 +28,12 @@ rules:
 - apiGroups: ["apiextensions.k8s.io"]
   resources: ["customresourcedefinitions"]
   verbs: ["get", "list", "watch"]
+- apiGroups: ["access.smi-spec.io", "specs.smi-spec.io", "split.smi-spec.io"]
+  resources: ["*"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["config.openservicemesh.io"]
+  resources: ["meshconfigs"]
+  verbs: ["get", "list", "watch"]
```
johnsonshi, line 36: @SanyaKochhar I seem to recall @arnaud-tincelin and @Tatsinnit mentioning that the `deployment/aks-periscope.yaml` file is being deprecated. I was previously told to make `aks-periscope-role ClusterRole` changes in https://github.com/Azure/aks-periscope/blob/master/deployment/cluster-role.yaml.

@Tatsinnit can you confirm?

SanyaKochhar: I did add these rbac permissions to `cluster-role.yaml` originally and I agree, it seems to make most sense to have them there, but it seemed like the rbac permissions doesn't work then. i.e., the `kubectl get meshconfig -A` command returns 
> Failed to collect meshconfigs for mesh osm: run command in container: exit status 1

Seeing similar error messaging for the SMI collector currently. Not sure if I'm missing some additional connection? @arnaud-tincelin @Tatsinnit please advise

SanyaKochhar: Discussed this with @Tatsinnit yesterday. I'll be moving the meshconfig ClusterRole to cluster-role.yaml so it can be deployed using Kustomize. Still testing this -- running into some issues with Kustomize currently

SanyaKochhar: Fixed this, moved cluster roles to cluster-role.yaml and tested this change with kustomize

pkg/collector/osm_collector.go
```diff
@@ -252,10 +253,17 @@ func (collector *OsmCollector) collectGroundTruth(meshName string) {
 		log.Print(validatingWebhookConfig)
 	}
 
-	collector.data[meshName+"_all_resources_list"] = allResourcesList
-	collector.data[meshName+"_all_resources_configs"] = allResourcesConfigs
-	collector.data[meshName+"_mutating_webhook_configurations"] = mutationWebhookConfig
-	collector.data[meshName+"_validating_webhook_configurations"] = validatingWebhookConfig
+	meshConfig, err := utils.RunCommandOnContainer("kubectl", "get", "meshconfigs", "--all-namespaces", "-o", "json")
```
Tatsinnit, line 256: 📔  Please note: for future work on this, lets make sure latter after GA of your need, we should plan a move towards using the `Client-go` packages into this collector, that will add better test-ability as well. https://github.com/kubernetes/client-go

There are some recent refactors we have done for those reason, one of the sample implementation is in ARC Pod Container work or system-err collector.

Tatsinnit (changes requested):
💡 Thank you so much for this PR and pining us for review.

I have added few key comments, I assume you have test all this for `SMI\OSM` scenario and all works as expected. 

* If the `yaml` changes are specific to `SMIOSM` needs you might only want to keep them in your consuming tools.
* Please add some unit-test for this collector if possible, currently it decreases the over-all code-coverage. ( image )
* Regarding the over all, after you are GA, we should align a work for converting it using `client-go` package.

Thank you

pkg/collector/osm_collector.go
```diff
@@ -197,8 +198,8 @@ func (collector *OsmCollector) collectDataFromEnvoys(namespace string) error {
 			// Remove certificate secrets from Envoy config i.e., "inline_bytes" field from response
 			re := regexp.MustCompile("(?m)[\r\n]+^.*inline_bytes.*$")
 			secretRemovedResponse := re.ReplaceAllString(string(responseBody), "---redacted---")
-
-			collector.data[query+"_"+podName] = secretRemovedResponse
+			filePath := meshName + "/envoy/"+ podName + query
```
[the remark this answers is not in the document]
SanyaKochhar: There can be multiple meshes in a cluster and the collector collects all resources for each of them, so the number of files was getting overwhelming and had to discern between. Passing in `mesh` allows the file system to be organized so that resources collected for each mesh are grouped together, at least. Initially I was trying to just use a smarter naming system without creating subdirectories but the names were getting very long and unwieldy, ex: `<meshName>_namespace_<namespaceName>_pod_<podName>_logs`. If your team is open to it, I would prefer to include further subdirectories for each mesh, namespace, etc. as we had before the code was refactored so we can avoid such complex naming

cc: @arnaud-tincelin

pkg/collector/osm_collector.go
```diff
@@ -220,8 +221,8 @@ func (collector *OsmCollector) collectPodLogs(namespace string) error {
 			output = fmt.Sprintf("Failed to collect logs for pod %s: %+v", podName, err)
 			log.Print(output)
 		}
-
-		collector.data[podName+"_logs"] = output
+		filePath := meshName + "/" + podName + "_podLogs"
```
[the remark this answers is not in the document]
SanyaKochhar: yes, more context here: https://github.com/Azure/aks-periscope/pull/118#discussion_r714008429

pkg/collector/osm_collector.go
SanyaKochhar: Sure, that makes sense

deployment/cluster-role.yaml
```diff
@@ -12,6 +12,12 @@ rules:
 - apiGroups: ["apiextensions.k8s.io"]
   resources: ["customresourcedefinitions"]
   verbs: ["get", "list", "watch"]
+- apiGroups: ["admissionregistration.k8s.io"]
```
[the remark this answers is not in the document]
SanyaKochhar: I can move these out, I'm just wondering -- removing this rbac will mean that the SMI and OSM collectors will not completely work out of the box if someone were to try to use them. Is that something we want? An alternative would be to add some documentation around the kustomize changes required to use those two collectors

pkg/collector/osm_collector.go
SanyaKochhar: Added a test that I will also update once we refactor for client-go. Currently it's hard to test without client go.

Tatsinnit (changes requested):
💡 🙏 Thanks for this, there are few comments, especially the yaml changes are something which need re-look. I see mainly PR is just the name and folder structure change and if you have tested this and only thing left is the `yaml` changes which is not needed in the core.

Can you please open a tracking item for the conversion of these collector to `client-go` we have already implemented logic `clients` and `podDecribe` For example a sample reside here: https://github.com/Azure/aks-periscope/blob/master/pkg/collector/kubeobjects_collector.go#L54

for future purpose regarding the `client-go > api` based implementation, I can help and implement this the conversion with you but I will need someone who understand the SMI/OSM scenario to test them alongside. 🙏☕️

Thank you so much,

pkg/collector/osm_collector_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	appsv1 "k8s.io/api/apps/v1"
+)
+
+func TestOsmCollector(t *testing.T) {
+	tests := []struct {
+		name    string
+		want    int
+		wantErr bool
+		deployments []*appsv1.Deployment
```
Tatsinnit, line 16: 💡 Idea: if you add `collectorName` like in other tests you could test the `GetName()` as well.

so add `collectorName string` here then, below  `value` --> `collectorName "osm"` later under the the test just do this around `line 47`: 

```
        name := c.GetName()
	if name != tt.collectorName {
		t.Errorf("GetName()) = %v, want %v", name, tt.collectorName)
	}
```

deployment/cluster-role.yaml
```diff
@@ -12,6 +12,3 @@ rules:
 - apiGroups: ["apiextensions.k8s.io"]
   resources: ["customresourcedefinitions"]
   verbs: ["get", "list", "watch"]
-- apiGroups: ["access.smi-spec.io", "specs.smi-spec.io", "split.smi-spec.io"]
```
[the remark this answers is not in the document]
SanyaKochhar: It's SMI specific, which is why I have removed it from here

SanyaKochhar:
> 💡 🙏 Thanks for this, there are few comments, especially the yaml changes are something which need re-look. I see mainly PR is just the name and folder structure change and if you have tested this and only thing left is the `yaml` changes which is not needed in the core. Thank you for adding very initial test class for this as well.
> 
> Can you please open a tracking item for the conversion of these collector to `client-go` we have already implemented logic `clients` and `podDecribe` For example a sample reside here: https://github.com/Azure/aks-periscope/blob/master/pkg/collector/kubeobjects_collector.go#L54
> 
> for future purpose regarding the `client-go > api` based implementation, I can help and implement this the conversion with you but I will need someone who understand the SMI/OSM scenario to test them alongside. 🙏☕️
> 
> Thank you so much,

That is correct, I've now updated the title of the PR to reflect the same and opened an issue to track the conversion to client go here #124. Thanks for your help with this.

Tatsinnit (approved):
👍

PR #122, opened 2021-09-28
Possible fix for the dependency alert.

Tatsinnit:
Gave a quick try to fix the dependency alert regarding: `github.com/containerd/containerd` and `github.com/opencontainers/runc` cc: @arnaud-tincelin 

**Update**: so seems like this will **fix** the dependency issue: I tested by enabling that in my fork with same commit and it stopped showing dependency alerts: https://github.com/Tatsinnit/aks-periscope 

Note: These packages are not directly used and I believe this is probably used via `helm` go-lang packages.

private image: `docker.io/tatsat/dependency-alert` 

cc: @rzhang628 and @qpetraroia for visibility.

Thanks,


sophsoph321 (approved):
Tested and these changes do not break arc scenarios

Issue #134, opened 2021-11-14
Collect PDB related information.


Tatsinnit:
This is a feature request, since we collect various information as an upgrade it will be amazing to log the `PodDisruptionBudget` information as well.

PDB in itself is a huge topic, but a log which can give enough data for the end user who understand will help in further analysis. Collecting this information and making this available via this tool helps in CSS or anyone analysing rule out PDB related issues.

Extras PDB information: https://kubernetes.io/docs/tasks/run-application/configure-pdb/ 

cc: @SriHarsha001 - might like this from this tool, (This has no direct connection with the other diagnostic tool which is based on rules engine) but I thought it might be good to add this information as PDB logs.

Any thoughts? please feel free to share.

SriHarsha001:
@Tatsinnit - This is awesome. Yes, it makes complete sense to get a dump of PDB information from the cluster to analyze.

PR #135, opened 2021-11-17
add pod distribution budget information collector.

Tatsinnit:
This implementation is to collect the information which usually help folks for identification of Pod disruption budget related glitches.

This PR should take care of #134 , cc: @SriHarsha001 as he is aware of these logic essentially this is just raw collection of data which could help in identification of PDB related issues. Plus @bcho 

Thanks. 🙏

### DIFF
--- a/pkg/collector/pdb_collector.go
+++ b/pkg/collector/pdb_collector.go
@@ -49,7 +49,7 @@ func (collector *PDBCollector) Collect() error {
 	}
 
 	for _, namespace := range namespacesList.Items {
-		podDistInterface, err := clientset.PolicyV1beta1().PodDisruptionBudgets(namespace.Name).List(context.Background(), metav1.ListOptions{}) //.Get(context.Background(), namespace.Name, metav1.GetOptions{})
+		podDistInterface, err := clientset.PolicyV1beta1().PodDisruptionBudgets(namespace.Name).List(context.Background(), metav1.ListOptions{})
 		if err != nil {
 			return fmt.Errorf("PDB error cluster: %w", err)
 		}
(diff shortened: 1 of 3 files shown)


pkg/collector/pdb_collector.go
```diff
@@ -0,0 +1,81 @@
...
+
+// Collect implements the interface method
+func (collector *PDBCollector) Collect() error {
+	// Creates the clientset
+	clientset, err := kubernetes.NewForConfig(collector.kubeconfig)
+	if err != nil {
+		return fmt.Errorf("getting access to K8S failed: %w", err)
+	}
+
+	namespacesList, err := clientset.CoreV1().Namespaces().List(context.Background(), metav1.ListOptions{})
```
bcho, line 46: nit: we properly can create a context at the top of this function, then reuse it in the list ns / pdb calls

pkg/collector/pdb_collector.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to list namespaces in the cluster: %w", err)
+	}
+
+	for _, namespace := range namespacesList.Items {
+		podDistInterface, err := clientset.PolicyV1beta1().PodDisruptionBudgets(namespace.Name).List(context.Background(), metav1.ListOptions{}) //.Get(context.Background(), namespace.Name, metav1.GetOptions{})
```
bcho, line 52: I think we can remove the commented code?

peterbom (approved):

pkg/collector/pdb_collector.go
Tatsinnit: Thank you sounds good, 👍 lemme know how it looks now ☕️☕️

bcho (approved):

PR #136, opened 2021-12-06
Use client-go and remove unecessary kubectl.

Tatsinnit:
This PR is important for 2 reasons:

* This pull request removes leftover kubectl dependencies and the only leftovers are the `OSM` and `SMI` collectors only.
* This pull request moves the repo more closer to a goal of stop using the `kubectl` command and use `client-go` module instead.

The work also redo the `creationTime` fetch and removes bunch of stuff which seems like unnecessary and not in need, I will ping JunSUn to take a look but I have tested this agains the vanilla `AKS Cluster` and seems to be behaving fine.

* Test image: `docker.io/tatsat/misc-move-to-client-go` 

Thanks,

cc: @rzhang628 , @arnaud-tincelin, @JunSun17 , @sophsoph321 , plus @bcho ❤️🙏☕️

### DIFF
--- a/cmd/aks-periscope/aks-periscope.go
+++ b/cmd/aks-periscope/aks-periscope.go
@@ -16,7 +16,12 @@ import (
 )
 
 func main() {
-	creationTimeStamp, err := utils.GetCreationTimeStamp()
+	config, err := restclient.InClusterConfig()
+	if err != nil {
+		log.Fatalf("Cannot load kubeconfig: %v", err)
+	}
+
+	creationTimeStamp, err := utils.GetCreationTimeStamp(config)
 	if err != nil {
 		log.Fatalf("Failed to get creation timestamp: %v", err)
 	}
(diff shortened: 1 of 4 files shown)


cmd/aks-periscope/aks-periscope.go
```diff
@@ -26,10 +31,6 @@ func main() {
 		log.Fatalf("Failed to get the hostname on which AKS Periscope is running: %v", err)
 	}
 
-	if err := utils.CreateCRD(); err != nil {
```
bcho, line 29: question: will we still write and apply the crd?

Tatsinnit: 💡 Thank you so much for this question: Not sure why this code existed in first place, like we use `daemonset` to deploy against cluster, why do we need `createCRD` against a node and create a `diagnostic.yaml` file which seems to be doing nothing? 

hence I removed this, and with new image published as test `docker.io/tatsat/misc-move-to-client-go` I see no difference in the output of this tool, so I wonder what was the use of this `createCRD`.

bcho: Got it, thanks for the context

bcho (approved):

cmd/aks-periscope/aks-periscope.go
Tatsinnit: Thanks heaps to you bro, I think it is one of the very good question and I've wondered why this code existed, given its now 2 years, I wonder if there is more context to add @JunSun17 , but it works fine without it. ❤️🙏☕️

sophsoph321 (approved):
Changes look good for arc scenarios

Issue #30, opened 2020-07-17
Wait for underlying operation like upload to finish before returning successful message.


Tatsinnit:
There are many times I have noticed that the `aks-periscope` command ran successfully but actually the logs get uploaded in decent amount of say 10 to 5 second delay in there storage.

For example: I had more than 3 nodes in my cluster and then once the tool returned successful run status, I went in and checked since my storage explorer was already running in parallel and I had to come back and refresh in few seconds to see all my logs.

the downside of this tool returning success before actually the storage is being uploaded is that it gives false positives and that makes other tool to extend or adopt this tool since the results is not deterministic from the operation of running this command.

💡 Idea: Very likely some re-work is in dire need to fix this internal design level behaviour to reciprocate the difference between successful vs failed status.

Thanks,

Tatsinnit:
💡 This is debatable reason being, I am just adding thoughts from both sides I guess:

* The daemont set operation runs at the background and once the storage is all uploaded the things do appear in storage now this is by design and there is **No** upfront delay in the `apply` operation to finish.
* If we `wait` then it could be that `apply` might take some time which is adding `delay` to the current upfront behaviour. 

Just, adding it for more thoughts.

Thanks.

Tatsinnit:
💡 🧙  Some more information to keep this Issue more relevant for right fix to go in, essentially `blob.upload` is a function which is non-async function but to actually upload the data it does in batches and not all at once. Details here: https://docs.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-go?tabs=linux#upload-blobs-to-the-container 

Read these specific comments from the doc.

> // You can use the low-level Upload (PutBlob) API to upload files. Low-level APIs are simple wrappers for the Azure Storage REST APIs.
// Note that Upload can upload up to 256MB data in one shot. Details: https://docs.microsoft.com/rest/api/storageservices/put-blob
// To upload more than 256MB, use StageBlock (PutBlock) and CommitBlockList (PutBlockList) functions. 
// Following is commented out intentionally because we will instead use UploadFileToBlockBlob API to upload the blob
// _, err = blobURL.Upload(ctx, file, azblob.BlobHTTPHeaders{ContentType: "text/plain"}, azblob.Metadata{}, azblob.BlobAccessConditions{})
// handleErrors(err)

PR #137, opened 2021-12-08
Wait for all operations to finish.

Tatsinnit:
This pull request caters #30 and indirectly caters a issue which very rarely seen in some consuming tools like https://github.com/Azure/vscode-aks-tools/issues/44 

Just like collectors we have now added `wait` logic for the `zip` and `export to storage` functionality.

Happy to detail more if anyone keen, but I think the linked issues will help in understanding the idea behind this fix. Gist is that the operation of upload still happening at the background and by adding `wait` we are guaranteeing that this tool will return 100% success upload before it finishes, earlier the tool runs at the background due to `select {}` which leads to this kind of behaviour, as detailed: https://github.com/Azure/vscode-aks-tools/issues/44  its similar behaviour in az-cli as well.

* test image: `docker.io/tatsat/wait-for-all-operations`  

Thanks heaps,

cc: @rzhang628 , @arnaud-tincelin , @peterbom , @sophsoph321 , @bcho 🙏☕️❤️

### DIFF
--- a/cmd/aks-periscope/aks-periscope.go
+++ b/cmd/aks-periscope/aks-periscope.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"fmt"
 	"log"
 	"os"
 	"strings"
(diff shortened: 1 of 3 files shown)


sophsoph321:
Changes look good for arc scenarios

sophsoph321 (approved):

Tatsinnit:
Closing this PR in favour of #138 which fixes few things. Thanks, 🙏

PR #138, opened 2021-12-20
Behaviour fix, Upload API fix.

### LINKED ISSUE #30: Wait for underlying operation like upload to finish before returning successful message.

Tatsinnit:
This PR fixes few of long standing issues with `aks-periscope` tool.

* Successful upload before returning through blocking function, more details here: https://github.com/Azure/aks-periscope/issues/30 gist is : essentially `azblob.upload` is a function which is non-async function but to actually upload the data it does in batches and not all at once. Details here: https://docs.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-go?tabs=linux#upload-blobs-to-the-container .

* **Image to test**: `docker.io/tatsat/update-exporter`

Tested for Vanilla AKS scenario, and its cleaner now.

I am opening this PR and closing #137 in favour of these changes. This needs to be tested for ARC, and OSM/SMI as well, hence cc: @sophsoph321 , @SanyaKochhar , along with @arnaud-tincelin , @rzhang628 , @JunSun17, fyi: @peterbom and @bcho  - Thanks. 

Thanks heaps, ☕️🙏

### DIFF
--- a/pkg/exporter/azureblob_exporter.go
+++ b/pkg/exporter/azureblob_exporter.go
@@ -127,10 +127,9 @@ func (exporter *AzureBlobExporter) ExportReader(name string, reader io.ReadSeeke
 		return err
 	}
 
-	blob := containerURL.NewBlockBlobURL(fmt.Sprintf("%s/%s/%s", strings.Replace(exporter.creationTime, ":", "-", -1), exporter.hostname, name))
-	// _, err = blob.Upload(context.Background(), reader, azblob.BlobHTTPHeaders{}, azblob.Metadata{}, azblob.BlobAccessConditions{}, azblob.DefaultAccessTier, azblob.BlobTagsMap{}, azblob.ClientProvidedKeyOptions{})
+	blobUrl := containerURL.NewBlockBlobURL(fmt.Sprintf("%s/%s/%s", strings.Replace(exporter.creationTime, ":", "-", -1), exporter.hostname, name))
 	fmt.Printf("Uploading the file with blob name: %s\n", name)
-	_, err = azblob.UploadStreamToBlockBlob(context.Background(), reader, blob, azblob.UploadStreamToBlockBlobOptions{})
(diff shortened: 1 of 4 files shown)


pkg/exporter/azureblob_exporter.go
```diff
@@ -127,8 +127,9 @@ func (exporter *AzureBlobExporter) ExportReader(name string, reader io.ReadSeeke
 		return err
 	}
 
-	blob := containerURL.NewBlockBlobURL(fmt.Sprintf("%s/%s/%s", strings.Replace(exporter.creationTime, ":", "-", -1), exporter.hostname, name))
-	_, err = blob.Upload(context.Background(), reader, azblob.BlobHTTPHeaders{}, azblob.Metadata{}, azblob.BlobAccessConditions{})
+	blobUrl := containerURL.NewBlockBlobURL(fmt.Sprintf("%s/%s/%s", strings.Replace(exporter.creationTime, ":", "-", -1), exporter.hostname, name))
+	fmt.Printf("Uploading the file with blob name: %s\n", name)
```
bcho, line 131: nit: change to use `log.Printf`?

Tatsinnit: 👍  nice, thank you :) fixed.

bcho (approved):
LGTM

Tatsinnit:
💡 Update @sophsoph321 kindly tested this today and it worked. since she is away from keyboard I am updating that as archive. Thank you so much for test Sophie! ☕️🙏

PR #139, opened 2021-12-25
update v1beta1 apiextension to v1.

### LINKED ISSUE #141: CrashLoopBackOff aks-periscope pods with logs "failed to create CRD".
There is a specific behaviour I have noticed since last 2 days, which is detailed below. This is incidentally already fixed as an effort to move to use `client-go` hence for ARC and Vanilla AKS scenario we recently merged the PR to move away from the `kubectl direct command execution`. This issue seems like bubbled up because of the validation triggering now, otherwise the actual code existed from day-1.

After deploying aks-periscope tool, pods will `CrashLoopBackOff` with `error` and the log will look like this.

Following error:

```
❯ kubectl logs aks-periscope-cxbjk -n aks-periscope
2021/12/26 00:54:42 Failed to create CRD: run command in container: exit status 1
```

**Why this issue happening now & lesson to learn and move away from kubectl direct use**

* ~Not sure why this is happening since last 2 days, and I have no specific much answer to this but it could be that OS level `kubectl` binary and the direct call to `kubectl` is not working as it seems~.
     * **But**, I really feel this is a good example for us to learn and move away from direct use of `kubectl` in this tool, since `client-go` gives us better usability.

* I also feel that running `kubectl binary` was always a hack, which we have now transitioned all the collector minus OSM/SMI collectors to use `client-go` apis.

**Fix \ Solution**

This break seems like co-incidence but:

* Prima facie it looks like it has to do with a `CreateCRD` function which was always a questionable part and it was already removed as part of this PR #136 **as an effort to move away from kubectl binary**.

Logging this in case anyone else see same issue.

Thanks, ❤️🙏☕️

## Update 26 Dec 8:39 pm

Even though the culprit code is gone in newer version which I plan to release sometime soon, I added debugger and found this issue:

`2021/12/26 07:37:38 Failed to create CRD: run command: /usr/local/bin/kubectl apply -f aks-periscope-diagnostic-crd.yaml in container: exit status 1`.

so essentially this no longer works, and I am not sure possibly something within the cluster is improved to avoid this but I will keep an eye. The branch under my fork for this investigation is : `bug/investigation-Create-CRD` if anyone is keen.

Further logs: (This helps in explaining the issue possibly, now the part why it happened now is left.) 

```
❯ kubectl logs aks-periscope-xxxxxxx -n aks-periscope
2021/12/26 21:35:36 Failed to create CRD: run command: /usr/local/bin/kubectl apply -f aks-periscope-diagnostic-crd.yaml in container: exit status 1 and outputStream is - std out: , err: error: error validating "aks-periscope-diagnostic-crd.yaml": error validating data: ValidationError(Diagnostic.spec): unknown field "networkconfig" in com.github.azure.aks-periscope.v1.Diagnostic.spec; if you choose to ignore these errors, turn validation off with --validate=false
```

Tatsinnit:
This PR fixes the deprecation of `apiextensions.k8s.io/v1beta1` now moved to `v1`

* More details here: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122

This will avoid warnings like: `Warning: apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+;` 

Please note: ~this needs to be tested~. 

* Tested with: `docker.io/tatsat/deprecated-extension-crd-api`

Thanks. ☕️🙏

### DIFF
--- a/.github/workflows/ci-pipeline.yaml
+++ b/.github/workflows/ci-pipeline.yaml
@@ -7,7 +7,7 @@ on:
     branches: [ master ]
 
 env:
-  GO_VERSION: '1.15'
+  GO_VERSION: '1.16.1'
   KIND_VERSION: v0.11.1
   HELM_VERSION: v3.4.0
 
@@ -110,7 +110,7 @@ jobs:
           (cd ./deployment && kustomize edit set image aksrepos.azurecr.io/staging/aks-periscope=localhost:5000/periscope:foo)
           kubectl apply -f <(kustomize build ./deployment)
           kubectl -n aks-periscope describe ds aks-periscope
-          kubectl -n aks-periscope wait po --all --for condition=ready --timeout=60s
+          kubectl -n aks-periscope wait po --all --for condition=ready --timeout=240s
       - name: Go tests
         run: go test -v -race -coverprofile=coverage.out -covermode=atomic ./...
       - name: Upload coverage to Codecov
--- a/deployment/crd.yaml
+++ b/deployment/crd.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: diagnostics.aks-periscope.azure.github.com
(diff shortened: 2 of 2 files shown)


Tatsinnit:
☕️💡 hello reviewers 🙏 I am taking a /hallpass for this one since it’s holidays and seems like theee 2 files are not part of core, **but** required to help in fix for #141 , merging this now, thanks.

bcho (approved):

PR #140, opened 2021-12-25
Improve CI and add iptables and kubeletcmd test structure.

Tatsinnit:
Intent of this PR is to add the test structure for the `iptables` and the `kubeletcmd` test structure for consistency.

Thanks.

### DIFF
--- a/.github/workflows/ci-pipeline.yaml
+++ b/.github/workflows/ci-pipeline.yaml
@@ -98,7 +98,7 @@ jobs:
         with:
           context: .
           push: true
-          tags: localhost:5000/periscope:v1
+          tags: localhost:5000/periscope:${{ github.run_id }}
           file: ./builder/Dockerfile
       - name: Deploy dummy helm chart for tests
         run: |
@@ -107,7 +107,7 @@ jobs:
           helm install happy-panda bitnami/wordpress --namespace default
       - name: Deploy AKS Periscope
         run: |
-          (cd ./deployment && kustomize edit set image mcr.microsoft.com/aks/periscope=localhost:5000/periscope:v1)
+          (cd ./deployment && kustomize edit set image mcr.microsoft.com/aks/periscope=localhost:5000/periscope:${{ github.run_id }})
           kubectl apply -f <(kustomize build ./deployment)
           kubectl -n aks-periscope describe ds aks-periscope
           kubectl -n aks-periscope wait po --all --for condition=ready --timeout=240s
--- a/.github/workflows/ci-pipeline.yaml
+++ b/.github/workflows/ci-pipeline.yaml
@@ -110,7 +110,7 @@ jobs:
           (cd ./deployment && kustomize edit set image aksrepos.azurecr.io/staging/aks-periscope=localhost:5000/periscope:foo)
           kubectl apply -f <(kustomize build ./deployment)
           kubectl -n aks-periscope describe ds aks-periscope
-          kubectl -n aks-periscope wait po --all --for condition=ready --timeout=240s
+          kubectl -n aks-periscope wait po --all --for condition=ready --timeout=120s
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-pipeline.yaml
```diff
@@ -107,7 +107,7 @@ jobs:
           helm install happy-panda bitnami/wordpress --namespace default
       - name: Deploy AKS Periscope
         run: |
-          (cd ./deployment && kustomize edit set image aksrepos.azurecr.io/staging/aks-periscope=localhost:5000/periscope:foo)
+          (cd ./deployment && kustomize edit set image mcr.microsoft.com/aks/periscope=localhost:5000/periscope:v1)
```
bcho, line 110: question:  since kubelet will cache the image by tag, will it be better to tag the image with a unique name? Maybe we can use the `github.run_id` as tag here? https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

pkg/collector/iptables_collector_test.go
```diff
@@ -0,0 +1,38 @@
...
+			collectorName: "iptables",
+		},
+	}
+
+	c := NewIPTablesCollector()
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := c.Collect()
+			if (err != nil) == tt.wantErr {
```
bcho, line 27: nit/suggestion: though we already had many tests, will it be better to use testify? https://github.com/stretchr/testify We could save many boilerplate code with this package

pkg/collector/iptables_collector_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"testing"
+)
+
+func TestNewIPTablesCollector(t *testing.T) {
+	tests := []struct {
+		name          string
+		want          int
+		wantErr       bool
```
bcho, line 11: question: do we want to check the error signature as well?

pkg/collector/iptables_collector_test.go
Tatsinnit: ❤️❤️☕️ Thank you for this, 100 percent yes, yeah I think I will do another dedicated PR for that as a bulk update as one dedicated PR, I am also wishing/hoping of SMI and OSM moves to client-go use then we can pretty much delete the kubectl binary with image as well.

pkg/collector/iptables_collector_test.go
Tatsinnit: ❤️🙏 Thank you for asking this because this is great question for unlocking more improvement, Error handling throughout this repo is standard log.error usage, there is nothing in commonality to check. Again this is something we can discuss as thought share because I will be keen to pick this down for a dedicated “test improvement” PR.

Idea I have in mind and here is the share:

Upstream team has some way to create azure pipeline but I have no clue if they will allow to spin aks clusters if they do then we could use the whole end-to-end scenario for a pipeline. (This is all thoughts I don’t think they have any built-ins for aks cluster wipin ups and storage)

.github/workflows/ci-pipeline.yaml
Tatsinnit: Sounds good let me try it out and will push the change ☕️❤️

pkg/collector/iptables_collector_test.go
bcho: Got it, thanks for the context!

bcho (approved):
LGTM

.github/workflows/ci-pipeline.yaml
Tatsinnit: Please Note :) pushed these changes now. rest will take care in dedicated PR for test general health improvement \ refactors. ❤️☕️ Thanks you so much.

PR #142, opened 2022-01-03
fix api version in old yaml.

Tatsinnit:
Just to be safe lets fix the old `Yaml` until we get rid of it. 

The new way is already changed under this PR : https://github.com/Azure/aks-periscope/pull/139 

Thanks, 🙏☕️


rzhang628 (approved):
looks good to me, Approved.

Issue #163, opened 2022-04-05
Support and encourage use of Kustomize for consuming tools


peterbom:
**Suggested enhancement**

Some work has already been done to provide a `kustomize` base that can be used to ease Periscope deployment, and this is already used in the CI testing pipeline.

If consuming tools adopt this means of deployment it has the potential to:
- Avoid duplication of code by removing the entire Periscope yaml resource definition from those tools.
- Require much smaller and more targeted code changes (generally just release tag + image tag) whenever the Periscope resource specification changes, simplifying both updates and code reviews.
- Avoid magic string replacement (e.g. `"# <saskey, base64 encoded>"`).
- Make it easier to override configuration variables like `DIAGNOSTIC_NODELOGS_LIST` (the current mechanism requires grepping these values from the yaml).

There are a few changes we can make that can ease and encourage this approach. For example:
- Standardize the folder structure into [base + overlays](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/#bases-and-overlays).
- Provide one or more example overlays with documentation illustrating how they can be deployed, including supplying any additional configuration. A good candidate for this is for testing local changes in a Kind cluster during development.
- Include `secretGenerator` block in the base `kustomization.yaml`, so that this fully defines the required resources.
- Move `ConfigMap` resources into a `configMapGenerator` block in `kustomization.yaml`, for consistency with `secretGenerator`.
- Consolidate `configMapGenerator` resources so that a single `ConfigMap` is generated. This should simplify overriding config in overlays, as well as simplifying `DaemonSet` specifications (which will be helpful when we need to add another one for #10).
- Update the `README` to illustrate how we expect consumers to use Kustomize, in particular:
  - Referring to a tagged remote GitHub reference for a versioned, immutable base specification
  - Overriding configuration (diagnostic variables and storage settings)

PR #164, opened 2022-04-06
Support Kustomize for development and consuming tools

peterbom:
Addresses #163

Main changes:
- Shuffle resource deployment folder structure to be more consistent with standard Kustomize base/overlay layout
- Simplify image publishing workflows
   - building is done within Dockerfile, so it doesn't need repeating
   - remove extra tagging step
- Avoid editing kustomize base files in CI pipeline, so that
   - we no longer require the `kustomize` binary
   - we can run the same commands locally without overwriting source-controlled files
- Changes to the main `README.md` file to suggest alternative approaches for consuming tools
- Add some overlays (each with documentation describing its use):
   - `dev` for testing local changes in a `Kind` cluster
   - `dynamic-image` for running Periscope with arbitrary image/tags
   - `external` for generating deployment `yaml` files for consuming tools (until they can move to using Kustomize)


Tatsinnit (approved):
💡 👍 <3 Thank you for this PR, and leading this work - I will let you fly for this, Approving this one. This will open up windows support pathway which is so good.

Also we discussed about documentations et. al.

Issue #158, opened 2022-03-25
Node Logs Collector overwrites /var/log/azure/cluster-provision.log


peterbom:
**Describe the bug**
The default periscope configuration specifies that two node log files are saved:
- `/var/log/azure/cluster-provision.log`
- `/var/log/cloud-init.log`

But when you examine the output, you see that only the latter is saved.

**To Reproduce**
1. Deploy Periscope
2. Check uploaded logs under the `nodelogs` directory
3. View the listed files
4. There is just one file, called `nodeLog`, and it contains only the contents of `/var/log/cloud-init.log`

**Expected behavior**
There will be two files, named something like:
- `var_log_azure_cluster-provision.log`
- `var_log_cloud-init.log`
And that these contain the contents of the respective node log files

**Screenshots**
![nodelogs](https://user-images.githubusercontent.com/1817696/160035432-8d8353fa-abad-4816-8ff9-f054c28d38ea.jpg)

**Desktop (please complete the following information):**
 - OS: N/A
 - Version N/A

**Additional context**
This apparently slipped through the cracks on an earlier refactor. See: https://github.com/Azure/aks-periscope/pull/65/files#diff-c604028bb0caafbb4e42ce2c13ad12e129687dba8efacfaa604064b7fa1099d0

Reintroducing the following should fix the issue:
```
		normalizedNodeLog := strings.Replace(nodeLog, "/", "_", -1)
		if normalizedNodeLog[0] == '_' {
			normalizedNodeLog = normalizedNodeLog[1:]
		}
...
		collector.data[normalizedNodeLog] = output
```

PR #166, opened 2022-04-07
Fix Node Logs Collector to use separate keys for each log file

peterbom:
Addresses #158 - this creates a new key for each log file (sanitized by replacing '/' with '_') rather than writing every file to the same key.


Tatsinnit (approved):
👍🎉 Nice!! thank you for this, looks great! I think you have already tested these changes. We can merge this tomorrow.

Issue #10, opened 2019-09-27
Add Windows Support


PatrickLang:
There's no reason to have a separate project for Windows - merge it here.

/assign @PatrickLang

davidkydd:
Note there is an "equivalent" (for AKS anyway, for some version of equivalent) script for windows here: https://github.com/Azure/aks-engine/blob/master/scripts/collect-windows-logs.ps1

PR #167, opened 2022-04-11
Allow Periscope to run on Windows nodes

peterbom:
(Finally) addresses #10. Some notes:

- The publishing pipelines now create two images, 'aks/periscope' and 'aks/periscope-win'.
- This will require another MCR webhook (https://github.com/microsoft/mcr/pull/1833), which will need to be merged either:
  - before this PR; or
  - before the Build and Publish to MCR pipeline is next run.
- A new DaemonSet is required, because the `nodeSelector` and `volumes` property are at the DaemonSet level, not container level.
- There is more potential work to be done for Windows-specific diagnostics, see #10, but this at least allows most collectors to run.
- The main Periscope code change is to add a `CheckSupported` method to all collectors. The initial purpose of this was to check the OS type, but is intentionally generic to allow checking any contextual information (e.g. environment variables).
- Anything that uses `RunCommandOnHost` cannot be run on Windows (Windows does not support privileged mode).
- Code that used `cat` to get file contents also could not run on Windows.
- Getting the host name (which is essential for Periscope functionality) involved both `RunCommandOnHost` and `cat`, and so required a [completely different approach](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables).
- Anything else that used `cat` was replaced by host volume mapping and standard OS file reading (the volume mappings were already set up, so the `cat`s were unnecessary anyway).
- For all collectors that are _not_ supported on Windows, the reasons why are commented in the `CheckSupported` method.
- A lot of work went into making the code coverage checker happy and generally trying to make the code more testable.
  - Since ideally tests should run on both Linux and Windows, it is hard to test code that behaves conditionally based on `runtime.GOOS`.
  - This led to injecting a `RuntimeInfo` type into collectors, which would contain the OS identifier.
  - This led to putting _all_ environment variables into `RuntimeInfo`, making it easier to see what variables are being used, and removing the need to set env vars in test classes.
  - It was hard work figuring out what file paths and volume mappings were used, and adding extra Windows paths made this problem worse, so a `KnownFilePaths` type was added for carrying these around. As well as making file system usage more obvious and explicit, this allows a single type to be injected into collectors for more flexible testing.
  - Also due to differences between OS, tests that relied on hard-coded filesystem paths (e.g. the node logs collector) became inconsistent (there was already a TODO to that effect).
  - This led to injecting a `FileContentReader` interface into collectors, which could be substituted for a fake in-memory file content map.
  - The `FileContentReader` implementation is now moved out of `helper.go` into its own file, with unit tests. Unfortunately this made the coverage checker include the `utils` package, dramatically lowering overall test coverage, so I had to add an override to exclude it again.
  - Some of the logic in the main `aks-periscope.go` file involved conditionally adding collectors depending on environment variables. This kind of checking was a good fit for the `CheckSupported` method of the collectors themselves, and was moved there instead. This means (a) the main file is simplified, and (b) this logic is now completely covered by unit tests.

IMG_20220413_162456


Tatsinnit:
👍✅❤️🎉 Thank for this!! given an amount of thinking you have done around this Pete, I feel like you know more then what I can nit-pick and since size is decent, I can see that you have used client-go as baseline mechanism to implement underlying logic.

thank you so much for the diagram as well. I think you have plus one from my end for this sizeable PR. Just for sanity I might give another quick eyeballing tomorrow. 

Thank you!

Tatsinnit (approved):

PR #170, opened 2022-05-01
Make it easier to run and debug tests locally

### LINKED ISSUE #169: Make it easier to run and debug tests locally
**Motivation**
1. Tests run locally do not behave the same as the CI build, unless some quite elaborate manual steps are followed to exactly match the scripts defined in the CI workflow yaml.
2. Adding further test setup requires extra scripts in the CI workflow yaml, which is already quite extensive.

**Proposal**
We can define the test setup in the Periscope go code itself, using the following steps:
1. Use the golang docker client to create an `aks-periscope-test-tools` docker image that contains the setup tools (kubectl, kind, helm, etc.).
2. Use the docker client to run kind in a container (using the above image) to create the cluster.
3. Use the same approach to deploy resources (e.g., helm charts) to the cluster.

We could then:
1. Remove the setup scripts from the CI workflow, leaving the `go test ./...` command.
2. Run `go test` locally and expect the same behaviour.
3. Add a vscode `launch.json` config for debugging tests locally (again behaving exactly the same as the CI workflow).

**Benefits**
This will allow:
- Easier addition of further test setup, which can be run locally, allowing for a tighter dev loop.
- Easier increasing of test coverage, especially for OSM/SMI collectors. This in turn would allow:
   - converting these to client-go with greater confidence: #124 and #127; which would enable
   - running these to be run on Windows nodes: #168

peterbom:
**Changes**
- Use the golang docker client to create an aks-periscope-test-tools docker image that contains the setup tools (kubectl, kind, helm).
- Use the docker client to run kind in a container (using the above image) to create a Kind cluster.
- Use the same approach to deploy resources (e.g., helm charts) to the cluster.
- Remove the setup scripts from the CI workflow, leaving the go test ./... command.
- Add a vscode launch.json config for debugging tests locally (again behaving exactly the same as the CI workflow).
- Errorf string changes to keep the linter happy
- Increase the linter timeout to avoid false negative linter failures


Tatsinnit (approved):
💡🥷 Nice! @peterbom for this re-jigging the tests and intern making it more testable, also for the windows work which you have been doing. Thank you!! ❤️🙏☕️ 

LGTM, given the size of PR, there is nothing I can think of which needs attention so lets merge this and move forward, thanks .

PR #172, opened 2022-05-05
Document the automated testing approach introduced earlier

peterbom:
The previous changes to allow automated tests using a Kind cluster (#170) are lacking explanation of the rationale and mechanics of the approach. This is an attempt to rectify that.


Tatsinnit (approved):
👍

PR #173, opened 2022-05-06
Add test coverage for OSM and SMI collectors

### LINKED ISSUE #171: Need automated test coverage for OSM and SMI collectors
`OsmCollector` and `SmiCollector` currently have very little test coverage. It should now be possible to substantially increase that.

**Motivation**
We'd like to be able to refactor the OSM and SMI collectors to use client-go. This would allow:
- the removal of the kubectl binary from the image, reducing its size and perhaps allowing for a distroless image in the future
- removal of `RunCommandOnContainer` code
- making the OSM and SMI collectors available on Windows nodes

But before we can do that, we need adequate automated testing of these collectors to ensure our changes don't break their existing behaviour.

peterbom:
Addresses #171

This gives comprehensive test coverage for the happy path of the OSM and SMI collectors. The error handling parts that are not covered are fairly trivial and probably not worth the time investment required to get a high degree of coverage.

These collectors are no longer excluded from the coverage report. This actually causes a drop in reported coverage but is a more truthful metric.

The cluster configuration that's deployed in the setup for these tests is taken straight from the [OSM Bookstore Demo](https://release-v1-1.docs.openservicemesh.io/docs/getting_started/), with some minor tweaks to avoid unnecessary image pulls.

As well as exercising all the code involved in OSM and SMI collection, the content of the values collected is also validated by regular expressions. This will be important for verifying that future work to convert the collectors to use client-go preserves existing behaviour, e.g., the tests will fail if we inadvertently change a value from `yaml` to `json`, or report data for the wrong pod or namespace.

It also includes some refactoring of the cluster fixture setup code, like removing unnecessary public variables/functions from the `test` package and moving configuration variables to a known place.

cc. @SanyaKochhar for any thoughts/feedback.


Tatsinnit (approved):
👍 Looks good to me if you are happy, given the size, prima facie chaniges look great towards final moving away from `kubectl` binary dependency. Please let me know if there is anything specific, we need to discuss other than that I did ping @SanyaKochhar for a review. 

Given this PR is open for 25 + days lets go-ahead and merge this, I think this looks good, we can handle if anything latter.

Thank you.

PR #179, opened 2022-06-08
Use client-go for SMI collector

### LINKED ISSUE #127: Refactor SMI collector to use client-go.
Following workitem is associated with #124 in a sense that both the collectors extensively used the `kubectl` implementation which we should plan of converting into `golang` -> `pkg` based implementations which could involve using `kubernetes` - `client-go` or `kubernetes` or `apiextensionsClient` under the hood.

The long term benefit we are aiming for is the test-ability (Increase codeCoverage), along with removal of `kubectl binary` download, if we 100% move to using the pkgs, (which is medium to long term plan) 

some helpful pkgs: 

* https://github.com/kubernetes/client-go 
* https://github.com/kubernetes/apiextensions-apiserver 

Thanks heaps, fyi: @SanyaKochhar , thanks and I could collaborate with you, for help.

cc: @rzhang628 , @arnaud-tincelin , @qpetraroia

peterbom:
Addresses #127 

This is to take us one step further towards removing our reliance on the `kubectl` binary and parity between operating systems, making use of the `KubeCommandRunner` library dealing with `Unstructured` client-go types introduced in #178.

cc: @Tatsinnit, @SanyaKochhar

### DIFF
--- a/cmd/aks-periscope/aks-periscope.go
+++ b/cmd/aks-periscope/aks-periscope.go
@@ -60,7 +60,7 @@ func main() {
 		collector.NewOsmCollector(config, runtimeInfo),
 		collector.NewPDBCollector(config, runtimeInfo),
 		collector.NewPodsContainerLogsCollector(config, runtimeInfo),
-		collector.NewSmiCollector(runtimeInfo),
+		collector.NewSmiCollector(config, runtimeInfo),
 		collector.NewSystemLogsCollector(runtimeInfo),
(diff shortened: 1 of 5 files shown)


Tatsinnit (approved):
👍 /lgtm, thanks, deferring to @SanyaKochhar for an eye ball please. thank you so much.

SanyaKochhar (approved):

PR #181, opened 2022-06-11
Remove dependency on kubectl binary

peterbom:
Now that the OSM and SMI collectors are no longer dependent on `kubectl`, we can remove the binary from the Periscope image.

This removes it from the Docker image, along with other unused packages (taking the Linux image size from 161MB to 63MB).

It also updates the OSM and SMI collectors so that they will be enabled on Windows nodes.

To test manually, run Periscope from VS Code after setting the following in `settings.json`:

```json
    "aks.periscope.containerRegistry": "ghcr.io/peterbom",
    "aks.periscope.imageVersion": "0.0.9",
    "aks.periscope.releaseTag": "v0.9",
    "aks.periscope.repoOrg": "peterbom",
```

### DIFF
--- a/builder/Dockerfile.linux
+++ b/builder/Dockerfile.linux
@@ -20,11 +20,6 @@ RUN go build ./cmd/aks-periscope
 # Runner
 FROM alpine
 
-RUN apk --no-cache add ca-certificates curl openssl bash
-
-ADD https://storage.googleapis.com/kubernetes-release/release/v1.16.0/bin/linux/amd64/kubectl /usr/local/bin/kubectl
-RUN chmod +x /usr/local/bin/kubectl
-
 COPY --from=builder /build/aks-periscope /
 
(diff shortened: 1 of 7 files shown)


builder/Dockerfile.linux
```diff
@@ -20,11 +20,6 @@ RUN go build ./cmd/aks-periscope
 # Runner
 FROM alpine
 
-RUN apk --no-cache add ca-certificates curl openssl bash
```
Tatsinnit, line 23: 💡 we checked and this seems good, we dint find any code with curl dependency et. al. hence removal of this seems justifiable, if we find something we can always add it back again.

Tatsinnit (approved):

builder/Dockerfile.linux
Tatsinnit: Also, just as note, the default of `alpine` shell is `ash`

PR #182, opened 2022-06-13
Add required permissions for OSM and SMI collectors

peterbom:
This adds the required permissions for the OSM and SMI collectors to work fully.

~~Creating as draft for now, because I'm not sure whether these were omitted by design.~~

**EDIT:** Moving out of draft after confirming with @SanyaKochhar and @Tatsinnit that the Periscope service account indeed _should_ have these permissions.

~~If we _do_ want to make these changes, it might also be a good idea to run automated tests under an account which uses this role, so that any further missing permissions are discovered before merging.~~

**EDIT:** I made changes to the testing code so that it deploys the service account, cluster role and cluster role binding, and then uses the service account identity to connect to the cluster to run the tests. This also highlighted another existing bug; that is, the helm collector was also failing under the Periscope SA identity (this is now fixed by adding list-secrets permission).


Tatsinnit (approved):
👍 Nice, as long as we are keeping the current behaviour not broken, this change looks good, will defer to @SanyaKochhar for approval. thank you so much to both of you.

SanyaKochhar (approved):

PR #185, opened 2022-06-27
update containerd package.

Tatsinnit:
fixes: Dependency alerts, for old indirect use of containers package. thanks, cc @peterbom


peterbom:
lgtm

Issue #189, opened 2022-07-11
aks command invoke: evalsymlink failure on '/command-files/https:/github.com/


bhandariakshay1601:
Related command

az aks command invoke --resource-group <> --name <> --command "kubectl kustomize ." --file

Describe the bug
When trying to deploy aks periscope via aks command invoke, I am seeing the following error
Error: rawResources failed to read Resources: Load from path https://github.com/azure/aks-periscope//deployment/base?ref=v0.9 failed: evalsymlink failure on '/command-files/https:/github.com/azure/aks-periscope/deployment/base?ref=v0.9' : lstat /command-files/https:: no such file or directory

To Reproduce

If we run this command on any aks cluster
az aks command invoke --resource-group <> --name <> --command "kubectl kustomize ." --file

we will be able to reproduce
kustomization.YAML

resources:

https://github.com/azure/aks-periscope//deployment/base?ref=v0.9
images:

name: periscope-linux
newName: mcr.microsoft.com/aks/periscope
newTag: 0.0.9
name: periscope-windows
newName: mcr.microsoft.com/aks/periscope-win
newTag: 0.0.9
secretGenerator:

name: azureblob-secret
behavior: replace
literals:
AZURE_BLOB_ACCOUNT_NAME=<>
AZURE_BLOB_CONTAINER_NAME=<>
AZURE_BLOB_SAS_KEY=?<>
Expected behavior

COmmand should have been successful

Environment summary

Additional context

peterbom:
@bhandariakshay1601: thank you for logging the issue!

This looks very much like an incompatibility with the kubectl version (see [similar issue](https://github.com/Azure/aks-periscope/issues/184#issuecomment-1168192031)).

Would you be able to run `kubectl version` and upgrade if it's below 1.21? I'm hopeful that would resolve the error.

We have been intending to update our README to call out the version dependency, so thank you for the reminder!

Tatsinnit:
Thank you @peterbom for updating and taking a look, and thank you so much @bhandariakshay1601 for opening this, please let us know if there is anything else you want to be covered? otherwise the update of kubectl should fix the issue, thanks again

Tatsinnit:
💡Closing this issue for now, please let us know if there is anything else is in need and we can reopen this issue. Thank you so much all. 🙏☕️

PR #190, opened 2022-07-13
Add note to readme about kubectl version

peterbom:
Make it clear in the README that the kustomize deployment requires kubectl 1.21 or later.

See #184 and #189


Tatsinnit (approved):
☕️🎉👍

Issue #165, opened 2022-04-06
Periscope will only run correctly in the `aks-periscope` namespace


peterbom:
**Describe the bug**
Periscope cannot generate blob name for upload to storage if it is in a namespace other than `aks-periscope`. I'm not sure if this is a known and accepted limitation, but I couldn't find it documented and it tripped me up.

The reason for the hard-coded namespace dependency is:
- We need to generate a blob with a timestamp that's unique to the deployment/run, but shared between all pods.
- The method for getting this timestamp is to list all pods in the `aks-periscope` namespace, and take the creation timestamp of the last one.

Aside from tying us to a particular namespace, this is vulnerable to timing inconsistencies if an additional pod is created after another pod starts executing this code path. It also prevents us from addressing an issue raised in #157 (being unable to re-run Periscope for an existing DaemonSet). So we should probably be looking for a different approach anyway.

A bit of careful planning might enable us to address both this and #157 together. If the timestamp can be passed to periscope from something external to the pod (like the contents of a mounted file that can be watched), it could solve all of the above as well as giving us a potential means to trigger further 'runs' of an existing DaemonSet.

**To Reproduce**
1. Deploy using the usual yaml resource specification, but change the namespace to (e.g.) `aks-periscope-test`.
2. Wait for pods to be running
3. Check storage in the Azure Portal. Assuming everything else was set up correctly, the logs should have uploaded successfully, but the blob name will be `<no name>`.

**Expected behavior**
The blob name should be something like `2022-03-28T21-26-02Z`

**Screenshots**
<img width="242" alt="image" src="https://user-images.githubusercontent.com/1817696/162085683-16e001db-210c-402b-8871-2719fe51eccc.png">

**Desktop (please complete the following information):**
N/A

**Additional context**
N/A

peterbom:
This is related to an intermittent bug (which occurs more frequently on newer Windows clusters), in which the node folders in the output file structure are divided between _more than one_ timestamped root container.

By using the `DIAGNOSTIC_RUN_ID` variable now being supplied to Periscope by consuming tools, we can fix both of these bugs.

PR #191, opened 2022-07-21
Add Windows log collection collector

peterbom:
This adds a new collector that collates the results of running [collect-windows-logs.ps1](https://github.com/Azure/aks-engine/blob/master/scripts/collect-windows-logs.ps1) on the host node.

The collector itself doesn't actually run this script. It just detects when it has been run (for a particular deployment/run of Periscope) and gathers the log files. The script is run by pods in a separate DaemonSet, which has full access to the host environment (runs in a [host-process container](https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/)).

Why can't Periscope itself run in a host-process container? Theoretically it can. But due to the way these are implemented in upstream Kubernetes/containerd, the container image file-system is mounted at a non-root location (from the perspective of the container processes). That means any absolute paths we use inside the container will no longer point where we expect them to. This breaks existing code (like the client-go in-cluster config retrieval), and will make it exceedingly difficult to verify we don't have subtle bugs in code paths performing file access that are infrequently executed.

**Required changes to consuming tools**
- Periscope will still work unchanged if deployed as it is today
- To take advantage of the new functionality, tools will need to add the following to their `kustomization.yaml`, for clusters running Kubernetes v1.23+:
```yaml
components:
- https://github.com/azure/aks-periscope//deployment/components/win-hpc?ref=v0.0.10

configMapGenerator:
- name: diagnostic-config
  behavior: merge
  literals:
  - DIAGNOSTIC_RUN_ID=<YYYY-MM-DDTHH-mm-ssZ>
```
(The `DIAGNOSTIC_RUN_ID` variable allows log export to be run several times on the same node without ambiguity about which Periscope DaemonSet it is associated with. It will be possible to use the same mechanism in future to address #165 and #157.)

**Other notable changes:**
- `FileContentReader` has been renamed to `FileSystem`, because:
  - it no longer actually reads files; it now returns a reader from one of its methods
  - it performs other file system functions like checking file existence/size and listing directory contents
- Collectors, DataProducers and Diagnosers all previously had a `GetData` method that returned a map of keys to string data values. The value types have all been changed to a `DataValue` interface exposing a function returning an `io.ReadCloser`. This was because holding all the Windows log contents in memory caused the application to run out of memory.


Tatsinnit (approved):
🎉❤️☕️🚀

PR #196, opened 2022-08-26
Allow multiple 'runs' for Periscope

### LINKED ISSUE #165: Periscope will only run correctly in the `aks-periscope` namespace

peterbom:
This addresses #157, in particular [this clarification](https://github.com/Azure/aks-periscope/discussions/157#discussioncomment-2377289). Rather than replacing the DaemonSet with some other resource type (which we've discussed extensively, but not found anything suitable), we make the existing DaemonSet behave more like a DaemonSet. I.e., rather than running once and doing nothing, it continues listening for a configuration change that triggers it to run again.

The specific configuration change it is looking for is the `DIAGNOSTIC_RUN_ID` setting. This is now what determines the name of the container that logs are exported to, so each time this value is changed, Periscope will run again and export to a new container. This means we can now remove the `select {}` 'hack' that was introduced to prevent the DaemonSet going into CrashLoopBackoff.

It also fixes #165, which describes a couple of bugs that were side-effects of Periscope trying to determine its 'run ID' by itself rather than having it specified in configuration.

Some implementation notes:
- Configuration (both ConfigMap and Secret resources) are now mounted as file volumes in the container, rather than environment variables. The reason for this is that [environment variables are not updated](https://stackoverflow.com/a/56380121) when the underlying ConfigMap variables change. For consistency (and because other settings such as SAS token will also change between runs), this was done for *all* configuration from ConfigMaps and Secrets.
- There is a new `FileContentWatcher` type which is responsible for detecting changes in these mounted volumes (currently it's only used to monitor the `DIAGNOSTIC_RUN_ID` value). This has a naive polling implementation because there are no cross-platform mechanisms like [inotify](https://en.wikipedia.org/wiki/Inotify), and the most prominent [golang library for cross-platform file watching](https://github.com/fsnotify/fsnotify) was [archived](https://github.com/fsnotify/fsnotify/issues/413) when I last checked (it has since been restored, but we can manage without the dependency). There is not much point in trying to respond immediately to file content changes, because [there can be a two-minute gap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically) between updating the ConfigMap and file content changing within the pod.
- The OS identifier is moved out of the other runtime info settings, because (a) it'll never change for a single pod, and (b) runtime info values now depend on file paths, which depend on OS. It is now based on a `OSIdentifier` type, rather than a `string`.
- The `helpers.go` file was cleaned up (there were a few unused functions in there).
- A new `GetContent` function was added to `helpers.go`, to avoid duplication of code that read strings from functions returning an `io.ReadCloser`.
- The linter was complaining about use of the deprecated `io/ioutil` package, so all usages have been replaced by the equivalent functions in `io`/`os`.


Tatsinnit (approved):
❤️🙏☕️ THankyou so much for adding me to the review, looks awesome and especially enabling the polling mechanism rather then empty run is awesome.

as we discussed we will updated one scenario where if user is able to duplicate the `Run_ID` that could lead to side effect and we can get over it by documenting it.

Thanks heaps,

PR #202, opened 2022-12-02
Update deployment notes for image registries to include ACR as well as GHCR

peterbom:
This attempts to address some shortcomings in the READMEs for deploying Periscope using external image registries.

Thanks to @bravebeaver for pointing out that you can use the `az aks update --attach-acr` command to link an AKS cluster to an ACR.

Being able to use ACR is really handy because it allows us to publish and test images from a local filesystem (without needing to push to GitHub).

Even when it's not possible to use the `attach-acr` approach (which requires subscription `Owner` role assignment), you can still deploy a secret containing ACR credentials, so that approach is also described here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,9 +227,11 @@ CGO_ENABLED=0 GOOS=linux go build -mod=mod github.com/Azure/aks-periscope/cmd/ak
 
 See [this guide](./docs/testing.md) for running automated tests in a CI or development environment.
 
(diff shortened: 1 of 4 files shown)


Tatsinnit (approved):
❤️🎊☕️ Thanks for adding me into this review, looks good, I hope we have tested that vanilla scenario is not impacted form the `kustomise` file change, rest looks great! congratulations 🎉 both.

peterbom:
> ❤️🎊☕️ Thanks for adding me into this review, looks good, I hope we have tested that vanilla scenario is not impacted form the `kustomise` file change, rest looks great! congratulations 🎉 both.

Thanks for taking a look, @Tatsinnit! The `yaml` file change is for manual testing during development, and isn't used in any production scenarios, so we should be good on that front.

bravebeaver (approved):

PR #205, opened 2023-01-09
Go version upgrade.

bravebeaver:
- upgrade go to 1.19
- bump helm version


Tatsinnit (commented):
☕️🙏💡 suggestion / idea share: since we are updating the go-version let’s please consider updating to latest go version https://go.dev/doc/go1.19

thanks 🙏

.github/workflows/ci-pipeline.yaml
```diff
@@ -8,7 +8,7 @@ on:
     branches: [ master ]
 
 env:
-  GO_VERSION: '1.16.1'
+  GO_VERSION: '1.18'
```
peterbom, line 11: I see that this previously included the minor version. Any thoughts on whether we should be continuing to do that (e.g. 1.18.9, or 1.19.4 if we're going for the latest stable one)? I'd guess it'd make the CI builds more consistent.

bravebeaver: convinced. stability over "main version tag". will update

Tatsinnit:
💡 Thanks for this, I noticed something so mentioning : we can open PR from your fork - feature branch, why have we this `go1.18` from azure org level branch PR?

https://github.com/Azure/aks-periscope/tree/go1.18

Screenshot 2023-01-11 at 12 31 28 PM

bravebeaver:
close this one as a new PR is created following the correct fork model. thanks @Tatsinnit for pointing it out.

PR #206, opened 2023-01-09
Update codeql to latest versions and use.

Tatsinnit:
Updating CodeQL file as this was not updated for a bit and new versions are released. thanks in advance ❤️☕️💡 cc: @peterbom


peterbom (approved):
Thanks! I hadn't even _thought_ to keep this up to date... Looks good to me.

PR #209, opened 2023-01-11
Ig dns tracer

bravebeaver:
initial commit with unit test cases for dns tracer


pkg/collector/gadgets/dns_tracer_collector.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+}
+
+func (collector *DNSTracerCollector) GetName() string {
+	return "dns-tracer"
+}
+
+func (collector *DNSTracerCollector) CheckSupported() error {
+	// check for OS since ebpf is only available on linux
+	if collector.osIdentifier != "linux" {
```
Tatsinnit, line 36: 💡 please refer to this we could use `utils` for `linux or windows`, for conditional check instead of string comparison: https://github.com/Azure/aks-periscope/blob/master/pkg/collector/dns_collector.go#L39 


```*.go
if collector.osIdentifier != utils.Linux { ... }
```

pkg/collector/gadgets/dns_tracer_collector.go
```diff
@@ -0,0 +1,96 @@
...
+	defer func() {
+		if err := dnsTracer.Detach(pid); err != nil {
+			log.Fatalf("Could not detach tracer %v", err)
+		}
+	}()
+
+	//capture the result for 30 seconds.
+	done := make(chan struct{})
+	go func() {
+		time.Sleep(30 * time.Second)
```
Tatsinnit, line 86: 💡 Note: so in this functionality we capture on 30 seconds of information? (that's why this `sleep` is in place?)

pkg/collector/gadgets/dns_tracer_collector.go
```diff
@@ -0,0 +1,96 @@
...
+		close(done)
+	}()
+
+	<-done
```
Tatsinnit, line 90: 💡Question: what is the use case scenario for the `done channel` use here please.

Tatsinnit (commented):
☕️💡Thanks for this PR. Added some basic review, sure Pete will take a look in details, thank you so much for this. 🙏🙏

Please note: if this is deliberate - the PR is from your local branch to `go 1.18` branch? Is it deliberate.

Second key observation: Why have we created this branch at `azure org level` please? https://github.com/Azure/aks-periscope/tree/go1.18  ; might be you can open PR directly from your fork feature branches?


Screenshot 2023-01-11 at 12 29 10 PM

deployment/overlays/dev/kustomization.yaml
```diff
@@ -15,12 +15,12 @@ patches:
   patch: |-
     - op: replace
       path: '/spec/template/spec/containers/0/imagePullPolicy'
-      value: Never
+      value: Always
```
peterbom, line 18: Not surprisingly, I think we've come up with different local development/testing workflows. It'd probably be good to put our heads together and see if we can come up with a process that works for both Windows/WSL and Mac systems.

The reason for the `imagePullPolicy: Never` here was that I was building images but _not_ pushing them to a remote registry. I was then loading them into a `kind` cluster so that they were available to the nodes. If the container runtime ever tried to pull the images it would obviously fail, because they didn't exist remotely.

While this workflow felt convenient to me, I've never worked on a Mac, and so there may be reasons I failed to think about why this wouldn't work for other people.

pkg/collector/gadgets/dns_tracer_collector.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// Create tracer. In this case no parameters are passed.
+	dnsTracer, err := tracer.NewTracer()
+	if err != nil {
+		return fmt.Errorf("error creating tracer: %s\n", err)
+	}
+	defer dnsTracer.Close()
+
+	pid := uint32(os.Getpid())
```
peterbom, line 71: This is a great start. I'm hoping we can make it behave the same as the DNS trace that you created in your other branch by using the IG daemonset.

To do that, I think we need to make use of a `ContainerCollection`, which will track all the pods/containers on the node, and dynamically include their PIDs in the trace. Let me know if you'd like to work together on that.

pkg/collector/gadgets/dns_tracer_collector.go
peterbom: Yep, that's why, @Tatsinnit. My only comment here is that for testing purposes I'd prefer to pass in the `duration` as an argument to the collector constructor.

pkg/collector/gadgets/dns_tracer_collector.go
peterbom: I'm probably missing something, but could the whole block be reduced to just:
```go
//capture the result for 30 seconds.
time.Sleep(30 * time.Second) // or time.sleep(collector.duration)
```

pkg/collector/gadgets/dns_tracer_collector_test.go
```diff
@@ -0,0 +1,80 @@
+package gadgets_test
```
peterbom, line 1: Is this a golang convention, to put tests in a different package? So far, I've just been creating `<filename>_test.go` in the same package. Is the goal to avoid accessing package internals from tests?

pkg/collector/gadgets/dns_tracer_collector.go
peterbom: Oh, just had another idea which might be better than passing `duration` to the collector. What about passing a `func` that takes a `done` channel? Then the main `aks-periscope` code in the `cmd` package can do something like:
```go
waiter := func(done chan struct) {
    time.Sleep(30 * time.Second)
    done<-nil
}
gadgets.NewDNSTracerCollector(osIdentifier, waiter)
```

That way we avoid race conditions in the tests in which we might execute some nslookup commands after the `Collect` function finishes (and we can make the function last only as long as it needs to, rather than the full 30s).

pkg/collector/gadgets/dns_tracer_collector.go
bravebeaver: 1. the PR is raised against the go1.18 branch so that the files changed in this feature are correctly highlighted. the intention is to merge the Go version update PR then we can adjust the base branch of this branch accordingly. sorry about the confusion

bravebeaver: 2. as discussed, I have forked the repo and am going to follow the proper workflow process. thanks @Tatsinnit for your generous help

bravebeaver: > 💡 please refer to this we could use `utils` for `linux or windows`, for conditional check instead of string comparison: https://github.com/Azure/aks-periscope/blob/master/pkg/collector/dns_collector.go#L39
> 
> ```go
> if collector.osIdentifier != utils.Linux { ... }
> ```

well spotted! updated.

pkg/collector/gadgets/dns_tracer_collector.go
bravebeaver: > Yep, that's why, @Tatsinnit. My only comment here is that for testing purposes I'd prefer to pass in the `duration` as an argument to the collector constructor.

totally agree with the weird approach to capture "live traffic" by setting a timeframe.  I will incorporate the configuration way with a reasonable default.

pkg/collector/gadgets/dns_tracer_collector.go
bravebeaver: probably i have not understood the purpose of the dns tracer in the periscope context. if only used as a  diagnostic tool, the dns trace job is done as soon as some dns activities are recorded? so ideally would be the job cancelled itself when it collects something with a timeout so that it is not holding up the entire aks-periscope process.

deployment/overlays/dev/kustomization.yaml
bravebeaver: It makes sense now and thanks for the clarification. the file change slipped into the commit accidentally.

my workflow consists of running local test on an external cluster, not locally hosted like kind or minikube. with `Never` it would not pick up updated images. 

i think a better approach for the [policy](https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy) might be to use `IfNotPresent` so that
- behaves like Never if using `kind` and image updated / exist through other channels,
- pull if not
- dev could update image tag (recommended) to force a new image to be pull from remote repo.

pkg/collector/gadgets/dns_tracer_collector.go
bravebeaver: absolutely!

bravebeaver:
relocate the branch to a[ forked repo](https://github.com/bravebeaver/aks-periscope/tree/feature/dns-tracer) following comments.

Issue #208, opened 2023-01-10
Periscope does not publish images for Windows Server 2022 nodes


sumicalbin:
**Describe the bug**
From the discussions I've been having it seems that this script, [collectlogs](https://github.com/microsoft/SDN/blob/master/Kubernetes/windows/debug/collectlogs.ps1), or even other logging scripts can be included in periscope. It is unclear whether this has been added as a customizable configuration on the yaml or is preset somewhere.
I am curious if this has been included and if there is an example for this?
**To Reproduce**
When I run periscope with the latest build, I do not see a collectlogs output.
The pod logs
![image](https://user-images.githubusercontent.com/102391206/212203269-eb6abe59-d9ea-47ed-893a-eb3ab4402cdc.png)

Part of my kustomization.yaml
![image](https://user-images.githubusercontent.com/102391206/212203204-72e1780b-aca8-44cb-a8a1-7bfee846fe08.png)

**Desktop (please complete the following information):**
 - OS: windows
 - Version: 2022

sumicalbin:
I also tried with `az aks kollect` 
my command `az aks kollect -g $myResourceGroup -n $myAKSCluster --storage-account $storageAccountName --sas-token $sasKey`
And the output that was printed
![image](https://user-images.githubusercontent.com/102391206/212206493-4261a945-c279-4e55-8f53-ca92e0c1c37d.png)

peterbom:
Hi @sumicalbin - thank you for reporting this. I think you've highlighted a few issues here.

First, running the collectlogs.ps1 script is currently not supported. Windows diagnostics uses the [collect-windows-logs.ps1](https://github.com/Azure/aks-engine/blob/master/scripts/collect-windows-logs.ps1) script. Is that not sufficient for your needs?

Periscope does allow collection of arbitrary log _files_ based on `DIAGNOSTIC_NODELOGS_LIST_WINDOWS` environment variable, but it doesn't allow users to configure scripts to run to generate those files. A workaround might be to run the script on the nodes yourself, then use the above env var to collect the resulting log files in Periscope. Would that be feasible?

Some other notes:
- I think the `ImagePullBackOff` for the pod running on the system node pool is because the `kustomization.yaml` doesn't specify the Linux image (for `name: periscope-linux`).
- I'm unsure why the Periscope pod on the Windows node is crashing (both when run manually and when using the `kollect` command). Is there any hint in the pod logs?
[...]

PR #212, opened 2023-01-25
Allow building and using images for Windows Server 2022 as well as 2019

peterbom:
This addresses #208.

To avoid needing different DaemonSets for each Windows version, the resource specs have been changed to use [multi-platform](https://docs.docker.com/build/building/multi-platform/) images. In other words, a single image tag refers to an image for whatever platform the container is deployed to.

This requires creating a manifest that associates the multi-platform image tag with each of the platform-specific tags, so both the GHCR and MCR workflows have been updated to publish those.

The Dockerfiles have also been updated to use lightweight golden MCR images as base images. The Linux version now uses the distroless Mariner base image, and the Windows versions are based on the lighter-wight Nanoserver, reducing image pulling time.

Note: the MCR workflow configuration (ACR and Azure app identity) is moved from GH _secrets_ to _variables_. This was due to some refactoring in which it made sense for job outputs to include the ACR, but that was failing because GitHub sensibly doesn't allow secret values to be passed around this way. It would've been possible to _just_ store the ACR name as a variable and keep the rest as secrets, but none of those values are really confidential and it's simpler to have them in the same place. It's probably worth having someone double-check my reasoning though. We'll have to update those variables in the Azure fork before the next MCR deployment.


builder/Dockerfile.windows
```diff
@@ -1,10 +1,12 @@
...
+
 # Builder
-FROM golang:1.18.0-windowsservercore-1809 AS builder
+# golang builder image is multi-platform
+FROM golang:1.19.5 AS builder
 
-ENV GO111MODULE=on \
-    CGO_ENABLED=0 \
-    GOOS=windows \
-    GOARCH=amd64
```
bravebeaver, line 7: just highlighting the diff, is GOARCH no long necessary? assume there is no support of windows 2019 on arm.

builder/Dockerfile.linux
```diff
@@ -1,11 +1,12 @@
+# Base image depends on target platform.
+# See: https://mcr.microsoft.com/en-us/product/cbl-mariner/distroless/base/about
+ARG BASE_IMAGE=mcr.microsoft.com/cbl-mariner/distroless/base:2.0
```
bravebeaver, line 3: love that the dockerfiles are consistent!

bravebeaver:
looks good to me!

bravebeaver (approved):

.github/workflows/build-and-publish.yml
```diff
@@ -7,49 +7,88 @@ on: [workflow_dispatch]
...
+      uses: mindsers/changelog-reader-action@v2
+      with:
+        validation_depth: 10
+        path: ./CHANGELOG.md
+    - name: Get Version Info
+      id: read_metadata
+      run: |
+        echo "Version: ${{ steps.changelog_reader.outputs.version }}"
+        echo "Changes: ${{ steps.changelog_reader.outputs.changes }}"
+        $owner = "${{ github.repository_owner }}".ToLower()
```
Tatsinnit, line 28: 💡observation/question - In bash Assignment of variables with space after the (=) sign, noticed spaces here  `$owner = "${{ github.repository_owner }}".ToLower()` - Is it because now we use `"PowerShell --> under ubuntu-latest"`?

Tatsinnit, line 28: 💡☕️ Idea only: Thinking out loud and sharing, shall we make this `Env Var` https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow

.github/workflows/build-publish-mcr.yml
```diff
@@ -6,63 +6,107 @@ permissions:
...
+        client-id: ${{ vars.AZURE_CLIENT_ID }}
+        tenant-id: ${{ vars.AZURE_TENANT_ID }}
+        subscription-id: ${{ vars.AZURE_SUBSCRIPTION_ID }}
+    - name: Publish cross-platform manifest
+      id: publish
+      run: |
+        $tagbase = "${{ needs.common.outputs.tagbase }}"
+        az acr login -n ${{ vars.AZURE_REGISTRY_SERVER }}
+        docker manifest create $tagbase ${{ needs.publish.outputs.linux }} ${{ needs.publish.outputs.win2019 }} ${{ needs.publish.outputs.win2022 }}
+        docker manifest push $tagbase
```
Tatsinnit, line 112: 💡 for historic Archival reasons: this command is experimental for now: https://docs.docker.com/engine/reference/commandline/manifest_push/ 

Screenshot 2023-01-25 at 2 01 08 PM

Tatsinnit (approved):
❤️🙏☕️ Thank you so much for adding me to this review, looks good, and I hope you have tested this from your repo :) 🐈🐈 which seems like the only way for sanity check.

Had some thoughts, idea, and information to share hence commented. Rest looks good to me. Thanks.

builder/Dockerfile.windows
peterbom: These are already set in the whichever base builder image is referenced in the Dockerfile(s), which should be appropriate for the platform on which the build happens, e.g.:
```bash
docker run --rm golang:1.9.5 go env
GOARCH="amd64"
...
GOOS="linux"
```

builder/Dockerfile.linux
peterbom: They were close enough that I consolidated them into one for a while. But eventually changed my mind and separated them again (they were just different enough for the special cases to be hard to reason about).

.github/workflows/build-and-publish.yml
peterbom: Yeah, exactly. These workflows have jobs that run on different platforms, and using PowerShell is the simplest way to make the scripting syntax the same for all of them. For consistency, I thought it'd be best for _all_ the jobs to use PowerShell, even if they didn't need to.

Also, although PowerShell doesn't care whether there's a space or not, I quite like having it there because it makes it (slightly) more obvious it's not bash, and might save us in future from adding bash syntax and wondering why it isn't working...

peterbom: In this case I probably wouldn't bother, because it's just local to that particular step, and not shared anywhere else. But that's a good point - I didn't think to consider how we might be able to improve our workflow files with env vars.

.github/workflows/build-publish-mcr.yml
peterbom: Interesting! I didn't notice that. I used it because it was documented [here](https://learn.microsoft.com/en-us/azure/container-registry/push-multi-architecture-images#push-a-multi-arch-image) as the recommended way to publish multi-arch images. I'm prepared to take a punt on this functionality not being removed, and handle it if the command/syntax changes.

PR #221, opened 2023-03-13
Add nsenter to Linux Periscope image

peterbom:
This fixes a bug introduced with the recent change to use distroless images, which don't contain the `nsenter` binary. This utility is still required to run processes on the host VM, and is used by the `iptables`, kubelet-command and system-logs collectors, and so those collectors are currently failing.

Note: Windows images are unaffected, since they never relied on `nsenter` and those collectors are disabled on Windows.

### DIFF
--- a/builder/Dockerfile.linux
+++ b/builder/Dockerfile.linux
@@ -18,9 +18,22 @@ COPY . .
 
 RUN go build ./cmd/aks-periscope
 
+# Add dependencies for building nsenter
+RUN apt-get update && \
+    apt-get install -y autoconf autopoint bison gettext libtool
+
+# Create a statically-compiled nsenter binary (see: https://github.com/alexei-led/nsenter/blob/master/Dockerfile)
+# nsenter versions: https://www.kernel.org/pub/linux/utils/util-linux/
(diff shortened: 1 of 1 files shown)


builder/Dockerfile.linux
```diff
@@ -18,9 +18,22 @@ COPY . .
 
 RUN go build ./cmd/aks-periscope
 
+# Add dependencies for building nsenter
+RUN apt-get update && \
+    apt-get install -y autoconf autopoint bison gettext libtool
+
+# Create a statically-compiled nsenter binary (see: https://github.com/alexei-led/nsenter/blob/master/Dockerfile)
+# nsenter versions: https://www.kernel.org/pub/linux/utils/util-linux/
+ADD https://github.com/util-linux/util-linux/archive/v2.38.tar.gz .
```
Tatsinnit, line 27: 💡☕️:+1: Seems like a sane choice, even though this was released last year.

Tatsinnit (approved):
☕️💡🥷

builder/Dockerfile.linux
peterbom: It looked to me as if there was only one more recent release, 2.39, but that was still at RC stage... Anyway, tested and it seems to work!

Issue #219, opened 2023-03-10
Memory requested too high in aks-periscope daemonSet 


kuzhao:
**Describe the bug**
[Not a bug but a potential point of improvement]
```
$ az aks kollect -n xxx -g xxx --kube-objects 'default/deployment/ratings-v1' --storage-account xxx
...
$ kubectl describe -n aks-periscope ds aks-periscope
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: aks-periscope
...
          resources:
            limits:
              cpu: '1'
              memory: 2000Mi
            requests:
              cpu: 250m
              memory: 500Mi      *******<<<<
...
```
From AKS user's practical perspective, this amount of ram request is too much for a debugging feature/plugin. Consumers like omsagent and osm-controller cost around 250Mi, which is still considered high.
**To Reproduce**
Deploy periscope through az aks kollect (in preview).

**Expected behavior**
Consider:
* Use less sub processes while collecting things sequentially
* The debug deploy does not have to be a ds.

peterbom:
Hi @kuzhao - thank you for highlighting this. We'll do some investigation to see how much memory we actually need, as I suspect you're right that it could be significantly lower.

Quick question: what's the 'debug deploy' you're referring to in your last bullet point?

kuzhao:
> Hi @kuzhao - thank you for highlighting this. We'll do some investigation to see how much memory we actually need, as I suspect you're right that it could be significantly lower.
> 
> Quick question: what's the 'debug deploy' you're referring to in your last bullet point?

Appreciate the attention, Peter. For "debug deploy", i might be using a fumble word imao...  im just referring to the aks-periscope (deployed via az aks kollect for debugging, the source of 'debug deploy').  
To expand more on this bullet, the periscope daemonSet runs on all nodes, although I set a specific k8s object to debug. Ideally, periscope should launch pod and collect only on nodes where the pods of target k8s object is running. 
The alternative could be either rs with podAffinity or ds with nodeSelector.

Tatsinnit:
> If that is a need for you, feel free to submit a feature request issue. It's always good to understand people's use cases. @Tatsinnit might recall if that's ever been discussed before?

Thank you so much for this discussion to you both ❤️ 🙏 Thanks @peterbom for the question and sorry for delayed reply, to the question the answer is --> no. we had no plan for this in initial days of this tool. Day-1 we were focused on broadcast mechanism hence the use of `daemonset` to run in all nodes.  

[...]

PR #222, opened 2023-03-13
Update resource requests and limits for all Periscope pods

peterbom:
This addresses #219.

The values are based on the actual pod memory (from metrics server) on an AKS cluster (incorporating a buffer to allow for some future extra memory usage or clusters where there is more data to process).

The Windows 'diagnostics collection' DaemonSet did not already have a resources block, so this is added here. The numbers are extremely high for this because the PowerShell script that does the log collection is very memory-hungry when compressing events into a zip file.

This also fixes a minor bug in the Windows log collection script that would occasionally result in it running without a `run_id` value when Periscope was repeatedly deployed to the same cluster without deletion.

### DIFF
--- a/deployment/base/daemon-set.yaml
+++ b/deployment/base/daemon-set.yaml
@@ -41,10 +41,10 @@ spec:
           mountPath: /etchostlogs
         resources:
           requests:
-            memory: "500Mi"
-            cpu: "250m"
+            memory: "40Mi"
+            cpu: "1m"
           limits:
-            memory: "2000Mi"
+            memory: "500Mi"
             cpu: "1000m"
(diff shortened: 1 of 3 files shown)


Tatsinnit:
☕️💡🙏 Looks good to me, and thanks for adding me to the PR. Only one thing to archive is that current limits are now in `win-hoc` daemonset we have 2gigs of memory and hope that won't cause OOM kill behaviours, leaving this as a comment of our offline discussion.

I cannot think of anything else which needs to be archived here.

Thanks heaps, ❤️❤️🙏🥷

Tatsinnit (approved):

PR #226, opened 2023-04-21
Handle individual errors for entries in zip file

peterbom:
When generating the zip file for Windows logs, there are sometimes permission errors accessing a file `C:\AzureData\CustomDataSetupScript.log`. I'm not sure why that is, but one file shouldn't prevent the rest of the zip file from being generated. This makes sure that other data does get uploaded successfully, and records an error in the logs for individual entries that fail.


Tatsinnit (approved):
❤️☕️🎉

PR #230, opened 2023-05-23
Add Inspektor Gadget TCP and DNS traces

peterbom:
This adds two collectors for Inspektor Gadget data: TCP and DNS traces (30 seconds each, with no namespace/node/pod filtering), building on initial integration work by @bravebeaver.

These collectors run in-process, meaning they invoke the Inspektor Gadget code directly, and do not install any other resources on the cluster.

This necessitated some changes to the way we launch tests (both locally from VS Code, and from the CI build). Because the tests are really E2E tests rather than unit tests, we wanted to actually run the Inspektor Gadget code from tests. And because running eBPF code requires elevated privileges, the tests now run as root.

cc: @Tatsinnit


.github/workflows/ci-pipeline.yaml
```diff
@@ -48,6 +48,6 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: Go tests
-        run: go test -v -race -coverprofile=coverage.out -covermode=atomic ./...
+        run: sudo go test -v -race -coverprofile=coverage.out -covermode=atomic ./...
```
Tatsinnit, line 51: 💡Thank you for this. Archiving only: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#administrative-privileges - if anyone ponders `sudo` use.

minor small question is: Why is `sudo` needed?

.vscode/launch.json
```diff
@@ -8,22 +8,15 @@
...
-            "type": "go",
-            "request": "launch",
-            "program": "${fileDirname}",
-            "mode": "test",
-            "env": {},
-            "buildFlags": "-race",
-            "args": ["-test.v"],
-            "showLog": true
+            "showLog": true,
+            // https://github.com/golang/vscode-go/blob/master/docs/debugging.md#debug-a-package-test-as-root
```
Tatsinnit, line 16: Question please: 💡 What is the relevance of this `.vscode/launch` file with this PR? aka if end-users editor is something else - say - `vim` or anything else do we need some kind of small doc for fyi?

Tatsinnit (approved):
👍 Thank you so much for this, Looks good to me. Let's definitely test this and fyi to @rzhang628 around this as well. Thank you ❤️ 🙏 sorry for the delay.

PR #233, opened 2023-09-26
Ensure security vulnerabilities are patched in MCR images

peterbom:
This includes changes for:
1. A scheduled pipeline to scan and patch the latest published images in GHCR and ACR/MCR.
2. Extra steps in our MCR publishing pipeline to patch before publishing.
3. Merging of the GHCR and MCR publishing pipelines to ensure consistency between container registries, and because the build/scan/patch process requires images to be published to a public registry.

Example runs:
- [Build and Publish](https://github.com/peterbom/aks-periscope/actions/runs/6427188728)
- [Patch](https://github.com/peterbom/aks-periscope/actions/runs/6427287965)

Notes on _why_ we're making these particular changes:

**Should we create a pipeline that runs on a schedule, or on demand?** I think it makes more sense to run it on a schedule. Reason: any issues that we detect should be in our dependency tree, and will change when new vulnerabilities are discovered in those dependencies. We won't know when that happens, so nothing will prompt us to run the pipeline.

**Should that pipeline scan the image built from the latest code in master, or the latest published images?** The published images are the ones that actually matter, because they're being used, but it would still be useful to know if our code introduced a dependency with a vulnerability. For a pipeline that runs on a schedule, I think we should be checking the latest published images, because our code doesn't change on a regular schedule. (In theory we could patch older images too, but that would be more work and massively increase pipeline execution times, and we can add that in later if we really need it.)

**Should the pipeline just create a report and send some notification if there are vulnerabilities, and let us know so we can manually run a pipeline to patch the images? Or just go ahead and patch them?** I think it should go ahead and patch. Realistically, we're not going to analyze the report ourselves and have the knowledge to make good decisions. And even if there's a risk that the patch breaks our functionality, security is more important than functionality. We are keeping unpatched images available with the `'unpatched` tag suffix, just in case something gets broken.

**Can we test the patched images before publishing?** We have no automated tests for images, and I don't think it'd be trivial to create them. But if we add the patching to the GHCR publishing pipeline, we can manually check the published images from VS Code, which we should probably be doing before publishing to MCR anyway.


.github/workflows/build-publish.yml
```diff
@@ -0,0 +1,199 @@
+name: Build and Publish to GHCR and MCR
+
+on: [workflow_dispatch]
```
sozercan, line 3: why manual? would you want to auto trigger off when a tag is cut?

.github/workflows/build-publish.yml
```diff
@@ -0,0 +1,199 @@
...
+    # If this OS supports patching images using Copa, push the unpatched image to GHCR with the tag suffix '-unpatched'.
+    # Then, scan and patch this image, and re-tag it with the original tag (no suffix).
+    - name: Push Unpatched Image to GHCR
+      if: matrix.canpatch
+      run: |
+        docker tag ${{ steps.getref.outputs.ghcr_image_ref }} ${{ steps.getref.outputs.ghcr_image_ref }}-unpatched
+        docker push ${{ steps.getref.outputs.ghcr_image_ref }}-unpatched
+    - name: Generate Trivy Report
+      if: matrix.canpatch
+      uses: aquasecurity/trivy-action@41f05d9ecffa2ed3f1580af306000f734b733e54 # v0.11.2
```
sozercan, line 125: make sure to add dependabot if you are pinning otherwise these will be stale over time

.github/workflows/build-publish.yml
peterbom: I think this is just because of the way we've historically managed releases. The process involves manually updating the changelog, and then running this action which creates the release/tag as well as building and publishing the images. There might be a more streamlined approach (@Tatsinnit - anything to add/suggest?).

.github/workflows/build-publish.yml
peterbom: Just trying to work out what we'd need to enable for this to work. It seems dependabot is paused on this repo because we haven't merged its PRs for too long. So I guess we'd need to do that. But also maybe we need this?
image
cc: @Tatsinnit - in case you know more about this than me...

Tatsinnit: 💡 Oh, yeah. How about we create tracking issue for this repo, separately for enabling this for dependent bot, given all the good work under this PR can go ahead and we can enable that dependent bot alert post this PR merge? How does that feel to all. thank you so much for this. ❤️ cc: @sozercan + @peterbom

.github/workflows/build-publish.yml
Tatsinnit: All sounds good, just as moving forward solution +1, at the moment, given its just me or Pete are the maintainer, we do manual build and publish for release cadence. We can post this merge think carefully about what pros and cons are for publishing trigger, tag cut vs this way. I am not sure where I have read pros and cons, was wondering if we can have this discussion post this PR. Sharing this link as fyi https://keepachangelog.com/en/1.0.0/ 

What o you think guys?

peterbom: I reopened #132 to discuss this. Let's keep the current manual trigger and consider other options later.

.github/workflows/build-publish.yml
peterbom: Created #234 for this.

Tatsinnit (approved):

.github/workflows/build-publish.yml
```diff
@@ -0,0 +1,199 @@
...
+      id: vulncount
+      run: |
+        $report_file = "report.json"
+        cat $report_file
+        $vuln_count = jq '.Results[0].Vulnerabilities | length' $report_file
+        echo "vuln_count=$vuln_count" >> $env:GITHUB_OUTPUT
+    - name: Copa Action
+      if: matrix.canpatch && steps.vulncount.outputs.vuln_count != '0'
+      id: copa
+      uses: project-copacetic/copa-action@1eb86b0907bce48225b66dc9488c7d329c2d48a0 # v1.0.0
```
peterbom, line 144: Note: need to upgrade to 1.0.1 which adds a curl retry when getting the latest copa version.

peterbom: Updated, but now waiting for the next release of copa which will include https://github.com/project-copacetic/copacetic/pull/377

